PR #48, opened 2025-07-31
Message reply

timurgilfanov:
I would like to built a new message reply Application test. User open application and see a chat list. New message appears in one of the chat. Unread counter in chat list item appears with a 1 number and partial text of the message. Perform tap on this chat item and see full text of the message. Reply with an answer. Return to chat list and see that no unread message badge on chat preview.


timurgilfanov:
@claude write a commit message for squash and merge

Issue #97, opened 2025-10-02
Add settings screen


timurgilfanov:
For users to be able to edit they settings an profile add Settings screen that available to going to from Main screen through bottom tab navigation.

Settings screen should contain:
- Profile: picture, name, and action to edit them
- Settings: interface language, and ability to choose one between English and Germany.

PR #104, opened 2025-12-03
Implement view model for language change screen

timurgilfanov:
This PR implement view model layer of language change feature.

### Change of contract
Contract between view model and screen changed: 

#### LanguageUiState
Instead of `data class LanguageItem(val name: String, val value: UiLanguage)` use `UiLanguage`. `name` is part of representation and how to represent should be decided in component of the screen that do representation.
`selectedLanguage` become nullable to init OrbitMVI container with state.

Before:
```
data class LanguageUiState(
    val languages: ImmutableList<LanguageItem>,
    val selectedLanguage: LanguageItem,
)
```
After:
```
data class LanguageUiState(
    val languages: ImmutableList<UiLanguage>,
    val selectedLanguage: UiLanguage?,
)
```
#### LanguageSideEffects
* `LanguageNotChangedForAllDevices` side-effect removed because backend will not support this feature.
* Added `Unauthorized` side-effect to handle lost of authorization case. Screen should navigate user to login screen.
* Added `data class ChangeFailed(val error: ChangeLanguageRepositoryError)` so screen can handle all errors that can happened on language change.

Related to #97.

### DIFF
--- a/app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelStateTransitionsTest.kt
+++ b/app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelStateTransitionsTest.kt
@@ -26,22 +26,8 @@ import timur.gilfanov.messenger.ui.screen.user.LanguageViewModelTestFixtures.cre
 class LanguageViewModelStateTransitionsTest {
 
     /**
-     * Verifies that the `languages` list property maintains referential equality across
-     * state changes, ensuring only the `selectedLanguage` is updated while the list itself
-     * is reused.
-     *
-     * **Why this matters:**
-     * - **Performance**: Jetpack Compose uses referential equality for recomposition.
-     *   If `languages` never changes reference, UI elements depending on it won't recompose
-     *   unnecessarily
-     * - **Memory efficiency**: The immutable list is created once during initialization
-     *   and shared across all state instances
-     * - **Correctness**: Verifies proper state management where only changing data is updated
-     *   while static data remains constant
-     *
-     * **What would be wrong:**
-     * If each `state.copy(selectedLanguage = X)` created a new `languages` list,
-     * it would cause unnecessary recompositions and memory allocations.
+     * Verifies that the `languages` list maintains referential equality across state changes.
+     * This avoids unnecessary allocations and Compose recompositions.
      */
     @Test
     fun `languages list remains constant throughout lifecycle`() = runTest {
@@ -80,23 +66,8 @@ class LanguageViewModelStateTransitionsTest {
     }
 
     /**
-     * Verifies that each state update creates a new [LanguageUiState] instance,
-     * ensuring state objects are never mutated in place and previous state instances
-     * remain unchanged.
-     *
-     * **Why this matters:**
-     * - **Correctness**: MVI pattern requires immutable state for predictable behavior
-     *   and unidirectional data flow
-     * - **Thread safety**: Immutable state can be safely shared across coroutines
-     *   without synchronization
-     * - **Time travel debugging**: Previous state instances must remain unchanged
-     *   for debugging tools to work correctly
-     * - **Compose integration**: Compose snapshot system relies on new object instances
-     *   to detect state changes and trigger recomposition
-     *
-     * **What would be wrong:**
-     * If state were mutated in place (`state.selectedLanguage = X`), it would break
-     * MVI guarantees, cause race conditions, and prevent Compose from detecting changes.
+     * Verifies that state updates create new instances, preserving previous state values.
+     * This ensures MVI immutability: state1 remains unchanged after state2 is emitted.
      */
     @Test
     fun `state immutability preserved data class copy creates new instance`() = runTest {


app/src/main/java/timur/gilfanov/messenger/data/source/remote/RemoteSettingsDataSourceNoop.kt
timurgilfanov-second: It will be nice to have real implementation of settings remote data source, so we can test networking. We can create fake server launched in Docker container to run tests against.

app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelObservationTest.kt
```diff
@@ -0,0 +1,179 @@
...
+        )
+        val settingsRepository = createSettingsRepositoryWithFlow(settingsFlow)
+        val viewModel = createViewModel(identityRepository, settingsRepository)
+
+        viewModel.test(this) {
+            val job = runOnCreate()
+
+            testScheduler.runCurrent()
+            testScheduler.advanceTimeBy(300)
+            expectNoItems()
```
timurgilfanov-second: It works without testScheduler calling. Unnecessary code must be removed.

```suggestion
            expectNoItems()
```

app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelObservationTest.kt
```diff
@@ -0,0 +1,179 @@
...
+import timur.gilfanov.messenger.ui.screen.user.LanguageViewModelTestFixtures.createSuccessfulIdentityRepository
+import timur.gilfanov.messenger.ui.screen.user.LanguageViewModelTestFixtures.createTestSettings
+import timur.gilfanov.messenger.ui.screen.user.LanguageViewModelTestFixtures.createViewModel
+
+@OptIn(kotlinx.coroutines.ExperimentalCoroutinesApi::class)
+@Category(Component::class)
+class LanguageViewModelObservationTest {
+
+    @get:Rule
+    val mainDispatcherRule = MainDispatcherRule()
```
timurgilfanov-second: We don't use Main dispatcher in LanguageViewModel and don't need to replace it with test dispatcher. Rule must be removed.

```suggestion
```

app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelObservationTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            expectState {
+                copy(selectedLanguage = UiLanguage.German)
+            }
+
+            job.cancelAndJoin()
+        }
+    }
+}
```
timurgilfanov-second: No change failure exists in test. Must add failure and then success update of setting value. View model must produce state with last setting value.

```suggestion
    fun `change failure does not break observation stream`() = runTest {
        val settingsFlow = MutableStateFlow<ResultWithError<Settings, GetSettingsRepositoryError>>(
            ResultWithError.Success(createTestSettings(UiLanguage.English)),
        )
        val identityRepository = createSuccessfulIdentityRepository()
        val settingsRepository = createSettingsRepositoryWithFlow(settingsFlow)
        val viewModel = createViewModel(identityRepository, settingsRepository)

        viewModel.test(this) {
            val job = runOnCreate()

            expectState {
                copy(selectedLanguage = UiLanguage.English)
            }
            
            settingsFlow.update {
                ResultWithError.Failure(GetSettingsRepositoryError.UnknownError(Exception("Test error")))
            }
            
            testScheduler.advanceTimeBy(300)
            expectNoItem()

            settingsFlow.update {
                ResultWithError.Success(createTestSettings(UiLanguage.German))
            }

            expectState {
                copy(selectedLanguage = UiLanguage.German)
            }

            job.cancelAndJoin()
        }
    }
}
```

app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelChangeLanguageTest.kt
```diff
@@ -0,0 +1,268 @@
...
+
+        viewModel.test(this) {
+            val job = runOnCreate()
+            val initialState = awaitState()
+            assertEquals(UiLanguage.English, initialState.selectedLanguage)
+
+            viewModel.changeLanguage(UiLanguage.English)
+            settingsFlow.update {
+                ResultWithError.Success(createTestSettings(UiLanguage.English))
+            }
```
timurgilfanov-second: Test completing the SUT's work. Let's use fake instead of stub for repository test double. It will connect changeUiLanguage() and observeSettings() over mutable setting flow inside repository.

```suggestion
            viewModel.changeLanguage(UiLanguage.English)
```

```
 class SettingsRepositoryFake(
      initialSettings: Settings,
      private var changeResult: ResultWithError<Unit, ChangeLanguageRepositoryError> =
          ResultWithError.Success(Unit),
  ) : SettingsRepository {

      private val _settingsFlow = MutableStateFlow(
          ResultWithError.Success(initialSettings)
      )

      override fun observeSettings(identity: Identity) = _settingsFlow

      override suspend fun changeUiLanguage(
          identity: Identity,
          language: UiLanguage
      ): ResultWithError<Unit, ChangeLanguageRepositoryError> {
          if (changeResult is ResultWithError.Success) {
              _settingsFlow.update { current ->
                  (current as? ResultWithError.Success)?.let {
                      ResultWithError.Success(it.data.copy(uiLanguage = language))
                  } ?: current
              }
          }
          return changeResult
      }
  }
```

app/src/test/java/timur/gilfanov/messenger/ui/screen/user/LanguageViewModelChangeLanguageTest.kt
```diff
@@ -0,0 +1,268 @@
...
+//
+//            viewModel.changeLanguage(UiLanguage.German)
+//
+//            val sideEffect = awaitSideEffect()
+//            assertIs<LanguageSideEffects.Unauthorized>(sideEffect)
+//
+//            job.cancelAndJoin()
+//        }
+//    }
+}
```
timurgilfanov-second: Nice to add IdentityRepository stub that first return success, and then error. Check that view model produce NotAuth side-effect. This test don't rely to change language. InitialStateTest is more appropriate for it.

app/src/main/java/timur/gilfanov/messenger/ui/screen/user/LanguageUiState.kt
```diff
@@ -11,6 +12,6 @@ import kotlinx.collections.immutable.ImmutableList
  * @property selectedLanguage The currently selected language
  */
 data class LanguageUiState(
```
timurgilfanov-second: Must add @Immutable or Compose may treat LanguageUiState as unstable, causing unnecessary recompositions every time the parent recomposes:
 1. UiLanguage is a sealed interface - Compose compiler doesn't automatically recognize it as stable
 2. ImmutableList from kotlinx-collections-immutable is NOT automatically recognized as stable by the Compose compiler

 Alternative: Configure stability globally in your Compose compiler options via a stability configuration file:

  // stability_config.conf
  kotlinx.collections.immutable.*

  This tells Compose that all kotlinx immutable collections are stable, reducing the need for manual annotations.

```suggestion
@Immutable
data class LanguageUiState(
```

timurgilfanov-second (changes requested):
### Must-fix
- [x] Add `@Immutable` to LanguageUiState
- [x] Update expectState at LanguageViewModelObservationTest.kt:42
- [x] Remove mainDispatcherRule
- [x] Remove testScheduler.runCurrent()
- [x] Use fake SettingsRepository in LanguageViewModelChangeLanguageTest.kt
- [x] Check state not changed after failures in LanguageViewModelChangeLanguageTest.kt
- [ ] Update test `change failure does not break observation stream` in LanguageViewModelObservationTest.kt
### Clarify
- [ ] "debugging tools to work correctly" in LanguageViewModelStateTransitionsTest.kt
### Nice-to-have
- [ ] Add test that check change failure and then change success leads to successful state update in LanguageViewModelChangeLanguageTest.kt.
- [ ] Add test that when IdentityRepository first return success, and then error view model produce NotAuth side-effect.
- [ ] Real implementation of settings remote data source

PR #167, opened 2026-02-04
Add architecture rules for UI state management

timurgilfanov:
This PR added two related rules for UI state management those applied to screens with overlapping async actions that required coordination to meet application requirements:
- Rule 1: Single authority for ordering rules and UI state commits
- Rule 2: Deterministic UI State Commits.

Together they lead us to solution that will be easy to change when business rules updates, took less time to debug and onboard new team member. Recommended pattern is MVI with actor and reducer and serialized processing.

### What **must** be resolved before merge
- [x] How this rules will be enforced? What can be enforced with CI and what needs a checklist item for PRs?

### Nice-to-haves before merge
- [x] Check existent code base for following rules in PR. Add follow-up issues for violations.

### Follow-up issues
- #168

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,7 +150,7 @@ This is an Android messenger application built with Kotlin and Jetpack Compose, 
   - Use cases make decisions based on sealed interface hierarchy, not cause inspection
   - Causes are extracted ONLY for logging and diagnostics
   - See `SyncAllPendingSettingsUseCase.kt` for reference implementation pattern
-- **MVI Store Pattern**: MVI model classes should ending with `Store` and must follow AR-01:
+- **MVI Store Pattern**: MVI model classes should end with `Store` and must follow AR-01:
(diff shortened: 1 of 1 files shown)


timurgilfanov-second (changes requested):
Orbit MVI is a valid implementation of MVI pattern with `intent` as `actor` and `reduce`. 

### Must-haves
- [x] Update `CLAUDE.md` to use Store end for models in MVI architecture.
- [x] Update `CLAUDE.md` to not insist on implementation pattern `asStateFlow`.
- [x] Exclude application of `NoMutableStateFlowExposureRule` and `ReduceOnlyInActorRule` for MVI model classes used Orbit, because Orbit enforces this rules by design.

### Nice-to-haves
- [ ] Add architecture rule that MVI models used Orbit must ends with Store
- [x] Add follow-up issues to rename MVI model classes to end with Store of ViewModel.

PR #206, opened 2026-02-16
feat(util): add `repeatOnSubscription` extension for `MutableStateFlow`

### LINKED ISSUE #190: Replace Orbit `repeatOnSubscription` for MVVM
## Issue
Starting observation of the flow unconditionally in init keeps these upstream flows active for the entire ViewModel lifetime, even when no UI is collecting state. In this app, MainScreen conditionally composes ChatListScreen by tab, so switching away from Chats removes the collector but does not immediately destroy the ViewModel; after #180 refactor, chat-list/update flows continue running in the background (unlike the previous `repeatOnSubscription` behavior), which causes avoidable background DB/work and battery cost.

With removing of Orbit from UI we need a replacement for `repeatOnSubscription`.

## Context

Orbit have the `repeatOnSubscription` function to cancel execution of the block if UI state don't have subscribers for some time, timeout is configurable:
```
public suspend fun repeatOnSubscription(
    block: suspend CoroutineScope.() -> Unit
) {
    coroutineScope {
        launch {
            containerContext.subscribedCounter.subscribed.mapLatest {
                if (it.isSubscribed) block() else null
            }.collect()
        }
    }
}
```
This helps to save resources and don't execute code if this results does nowhere.

## Recommended solution
Implement function `repeatOnSubscription` that depends on `_state : MutableStateFlow`.

### First draft
This draft is class-wide, we need one function `repeatOnSubscription` to apply on each block of code (e.g., `observeProfile`, `observeChatList`).
```
private val _state = MutableStateFlow(ChatListScreenState(isLoading = true))
val state = _state.asStateFlow()

init {
    viewModelScope.launch {
        var job: Job? = null
        _state.subscriptionCount
            .map { it > 0 }
            .distinctUntilChanged()
            .collect { haveSubscribers ->
                if (haveSubscribers) {
                    job = launch {
                        observe()
                    }
                } else {
                    //todo delay for 5 seconds before cancel
                    job?.cancel()
                }
            }
    }
}

private suspend fun observe() = coroutineScope {
    launch { observeProfile() }
    launch { observeChatList() }
    launch { observeChatListUpdating() }
}
```

timurgilfanov:
Closes #190

## Summary

- `repeatOnSubscription` runs `block` while the flow has active collectors and cancels it after `stopTimeout` with no collectors
- **`block` must not complete normally** — if it returns before being cancelled it will not be restarted, even with active collectors (pinned by test, documented in KDoc)

## Test plan

- [ ] Block starts only when first subscriber appears
- [ ] Block stays alive while subscribers are present, including across time
- [ ] Block is not cancelled immediately when last subscriber leaves — respects `stopTimeout` grace period
- [ ] Block is cancelled after `stopTimeout` elapses with no subscribers
- [ ] Block keeps running if subscribers return within `stopTimeout`
- [ ] Block restarts if subscribers return after `stopTimeout`
- [ ] Block is cancelled when parent scope is cancelled

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/test/java/timur/gilfanov/messenger/util/RepeatOnSubscriptionUnitTest.kt
+++ b/app/src/test/java/timur/gilfanov/messenger/util/RepeatOnSubscriptionUnitTest.kt
@@ -26,6 +26,8 @@ class RepeatOnSubscriptionUnitTest {
 
     companion object {
         private val STOP_TIMEOUT = 1.seconds
+        private val WITHIN_TIMEOUT = 500.milliseconds
+        private val LONG_ADVANCE = 10.seconds
     }
 
     private fun TestScope.subscribe(flow: MutableStateFlow<Int>): Job =
@@ -35,7 +37,7 @@ class RepeatOnSubscriptionUnitTest {
 
     private fun TestScope.launchRepeatOnSubscription(
         flow: MutableStateFlow<Int>,
-        block: suspend CoroutineScope.() -> kotlin.Unit,
+        block: suspend CoroutineScope.() -> Unit,
     ): Job = backgroundScope.launch {
         flow.repeatOnSubscription(stopTimeout = STOP_TIMEOUT, block = block)
     }
@@ -72,7 +74,7 @@ class RepeatOnSubscriptionUnitTest {
         runCurrent()
         assertEquals(1, startCount.get())
 
-        advanceTimeBy(10.seconds)
+        advanceTimeBy(LONG_ADVANCE)
         assertEquals(1, startCount.get())
 
         sub.cancel()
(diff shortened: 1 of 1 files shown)


app/src/test/java/timur/gilfanov/messenger/util/RepeatOnSubscriptionUnitTest.kt
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    fun `block keeps running while subscribers are active`() = runTest {
+        val flow = MutableStateFlow(0)
+        val blockJob = CompletableDeferred<Job>()
+
+        launchRepeatOnSubscription(flow) {
+            blockJob.complete(coroutineContext[Job]!!)
+            awaitCancellation()
```
timurgilfanov: Why we need it here?

PR #342, opened 2026-04-11
fix(auth): clean up stale user-scoped settings and sync jobs

timurgilfanov:
Closes #330

## Summary
Clears user-scoped settings state and pending sync work when an authenticated session ends or rotates so a later login does not inherit stale local data.

## Changes
- Add `SettingsDao.deleteAllByUser(...)` and `LocalSettingsDataSource.deleteAllForUser(...)` with retry and error mapping for transient and permanent SQLite failures
- Tag one-time settings sync work with `UserScopeKey` and add scheduler support to cancel all jobs for the outgoing user
- Add `SettingsRepository.deleteUserData(...)` and call it from logout cleanup
- Extend auth session handling so cleanup also runs when refresh-token rotation changes the `UserScopeKey` or when refresh ends with session expiry
- Update DI, fakes, stubs, and repository and use case tests around cleanup ordering and failure handling

## Important Design Decision
Cleanup is best-effort and keyed by the outgoing `UserScopeKey`. The app only deletes user-scoped settings after the local auth session is successfully cleared, and it logs cleanup failures instead of blocking logout or session-expiry handling. Token refresh also triggers cleanup when refresh-token rotation changes the scope key, preventing orphaned settings rows and WorkManager jobs from surviving across sessions.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

Targeted coverage includes DAO deletion, local data source retry and error mapping, repository cleanup orchestration, logout cleanup, token-refresh cleanup, and auth repository state handling.

## Screenshots / Demo
Not applicable; no UI changes.

## Acceptance criteria
- [x] On logout, all WorkManager jobs scoped to the outgoing `UserScopeKey` are cancelled
- [x] On logout, all settings rows for the outgoing `UserScopeKey` are deleted from the local database
- [x] Re-login does not retain stale settings data or queued sync work from the previous session

## Breaking Changes
None.

## Follow-ups
- [ ] #343 — `SyncSettingWorker` may recreate deleted settings rows if it is mid-execution when `deleteUserData` runs (WorkManager cancellation is async; full fix requires cooperative `isStopped` guards in the worker)
- [ ] #344 — `SettingsSyncScheduler.cancelUserScopedJobs` returns `Unit`; typed error signaling would align it with the data-layer convention and remove the `runCatching` workaround in `SettingsRepositoryImpl`
- [ ] #345 — Settings cleanup is triggered imperatively at each write site that transitions to `Unauthenticated`; a reactive observer on `authState` would make any future path to `Unauthenticated` automatically trigger cleanup


app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncScheduler.kt
```diff
@@ -20,4 +20,13 @@ interface SettingsSyncScheduler {
...
     fun schedulePeriodicSync()
+
+    /**
+     * Cancel all pending and running WorkManager jobs that were scheduled for the given user.
+     *
+     * Jobs are identified by the tag equal to [UserScopeKey.key] that is added when scheduling
+     * via [scheduleSettingSync]. This should be called on logout so that no stale work runs
+     * for the outgoing user.
+     */
+    fun cancelUserScopedJobs(userKey: UserScopeKey)
```
timurgilfanov, line 31: Is it worth to support the case of cancellation of cancellation of jobs? We can at least log, or change a control flow with retry if cancellation of jobs is not done.
```suggestion
    suspend fun cancelUserScopedJobs(userKey: UserScopeKey) : ResultWithError<Unit, CancelJobsError>
```

timurgilfanov:
## CodeRabbit review response

Assessed all 7 comments. Summary below.

---

### C1 — `cancelUserScopedJobs` can throw before `deleteAllForUser` runs ✅ Fixed

Valid. Wrapped the call in `runCatching` with a logger call so any scheduler exception is caught, logged, and cleanup continues to the local deletion step regardless.

---

### C2 — `cancelAllWorkByTag()` is async; race with local deletion → Follow-up #343

Valid concern. However, awaiting the `Operation` only ensures the cancellation is registered in WorkManager's DB — it does not stop a currently-running worker. Full elimination requires the worker to check `isStopped` before each DB write. Since cleanup is explicitly best-effort, this is tracked in #343 for investigation.

---

### C3 — `SettingsRepositoryFake.deleteUserData` doesn't mutate state on success ✅ Fixed

Valid. Updated to emit `SettingsResetToDefaults` from `settingsFlow` on success, matching the `changeUiLanguage` pattern. Tests that observe settings after a successful delete will now see the deletion reflected.

---

### C4 — Completed plan doc has stale ordering description ⛔ No action

Out of scope. `docs/plans/completed/` are historical planning artifacts. Correcting the sequencing in an archived doc provides no production or test correctness benefit.

---

### C5 — `restoreJob` not completed in `finally`; `authState.first()` hangs on unexpected throws ✅ Fixed

Valid. Wrapped `restoreAuthState()` in `try/finally` so `restoreJob.complete(Unit)` is guaranteed to run regardless of exceptions, preventing any `authState.first()` caller from hanging.

---

### C6 — Inline comment in `AuthModule.kt` violates KDoc-only convention ✅ Fixed

Valid. Removed the inline `//` block and moved the DI-cycle rationale into a `/** ... */` KDoc above `provideTokenRefreshUseCase`.

---

### C7 — Scope key derived from `AuthState.Authenticated`; cleanup skipped if state is Unauthenticated ⛔ No action

Edge case only. If `restoreAuthState()` throws (now guarded by C5's fix), auth state stays `Unauthenticated` and the scope key is unknown — cleanup cannot run regardless. Deriving the scope key from the outgoing refresh token would require significant API changes to expose which token `authRepository.refreshToken()` used, and the practical impact is negligible.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImpl.kt
```diff
@@ -467,6 +469,26 @@ class SettingsRepositoryImpl @Inject constructor(
...
+                logErrorMapping(
+                    "deleteUserData",
+                    error,
+                    mapped,
+                    (localError as? LocalStorageError.UnknownError)?.cause,
+                )
+                mapped
+            },
+        )
+    }
```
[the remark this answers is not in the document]
timurgilfanov: Fixed. Wrapped `cancelUserScopedJobs(userKey)` in `runCatching { ... }.onFailure { cause -> logger.e(...) }` so any exception from the scheduler is caught, logged, and does not prevent `localDataSource.deleteAllForUser(userKey)` from running.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncSchedulerImpl.kt
```diff
@@ -54,6 +55,10 @@ class SettingsSyncSchedulerImpl @Inject constructor(private val workManager: Wor
         )
     }
 
+    override fun cancelUserScopedJobs(userKey: UserScopeKey) {
+        workManager.cancelAllWorkByTag(userKey.key)
+    }
```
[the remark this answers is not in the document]
timurgilfanov: Valid concern. However, awaiting the `Operation` only guarantees the cancellation is registered in WorkManager's internal DB — it does not stop a currently-running worker. A mid-execution `SyncSettingWorker` can still write to the DB after `deleteAllForUser` completes. Full elimination requires the worker to check `isStopped` before each write (cooperative cancellation). Since `deleteUserData` is explicitly best-effort, this is tracked as a follow-up in #343.

core/domain/src/testFixtures/kotlin/timur/gilfanov/messenger/domain/usecase/settings/SettingsRepositoryFake.kt
```diff
@@ -58,4 +64,11 @@ class SettingsRepositoryFake(
         userKey: UserScopeKey,
     ): ResultWithError<Unit, SyncAllSettingsRepositoryError> =
         error("syncAllPendingSettings not configured for this test")
+
+    override suspend fun deleteUserData(
+        userKey: UserScopeKey,
+    ): ResultWithError<Unit, DeleteUserDataRepositoryError> {
+        lastDeleteUserDataKey = userKey
+        return deleteUserDataResult
+    }
```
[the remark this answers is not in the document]
timurgilfanov: Fixed. When `deleteUserDataResult` is `Success`, `deleteUserData` now calls `settingsFlow.update { ResultWithError.Failure(GetSettingsRepositoryError.SettingsResetToDefaults) }`, matching the `changeUiLanguage` pattern. Tests that observe settings after a successful delete will see the deletion reflected.

docs/plans/completed/2026-04-11-fix-stale-user-scoped-state-on-logout.md
```diff
@@ -0,0 +1,116 @@
...
+- Modify: `feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/domain/usecase/LogoutUseCaseImplTest.kt`
+- Modify: `app/src/test/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModelLogoutTest.kt`
+
+- [x] Add `SettingsRepository` constructor parameter to `LogoutUseCaseImpl` (between `authRepository` and `logger`)
+- [x] In `LogoutUseCaseImpl.invoke()`: read `authRepository.authState.first()`, extract `UserScopeKey` if `Authenticated`, call `settingsRepository.deleteUserData(userKey)`, log any failure, then call `authRepository.logout()`
+- [x] Update `AuthViewModelModule.provideLogoutUseCase` to inject `SettingsRepository`
+- [x] Update `LogoutUseCaseImplTest.createUseCase` to accept `SettingsRepository` (default: `SettingsRepositoryStub()`); add test: cleanup called with correct key on successful logout; add test: cleanup failure does not block logout
+- [x] Update `SettingsViewModelLogoutTest.createViewModel`: add `SettingsRepositoryStub()` to `LogoutUseCaseImpl` constructor
+- [x] Run `./gradlew :feature:auth:testDebugUnitTest` and `./gradlew app:testDebugUnitTest` — must pass
```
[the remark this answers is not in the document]
timurgilfanov: No action. `docs/plans/completed/` entries are historical planning artifacts; correcting the sequencing in an archived doc provides no production or test correctness benefit.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -49,11 +51,18 @@ class AuthRepositoryImpl @Inject constructor(
...
+        restoreJob.await()
+        emitAll(_authState)
+    }
 
     init {
-        coroutineScope.launch { restoreAuthState() }
+        coroutineScope.launch {
+            restoreAuthState()
+            restoreJob.complete(Unit)
+        }
```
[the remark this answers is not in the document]
timurgilfanov: Fixed. Wrapped `restoreAuthState()` in `try/finally` so `restoreJob.complete(Unit)` is guaranteed to run regardless of exceptions, preventing any `authState.first()` caller from hanging if restoration throws unexpectedly.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/di/AuthModule.kt
```diff
@@ -23,8 +28,36 @@ object AuthModule {
...
+    fun provideTokenRefreshUseCase(
+        authRepository: AuthRepository,
+        settingsRepositoryLazy: Lazy<SettingsRepository>,
+        logger: Logger,
+    ): TokenRefreshUseCase {
+        // Break the DI cycle:
+        //   TokenRefreshUseCase -> SettingsRepository -> RemoteSettingsDataSource
+        //   -> HttpClient -> AuthInterceptor -> TokenRefreshUseCase
+        // Dagger.Lazy defers SettingsRepository instantiation until first use, after
+        // all singletons in the cycle are already constructed.
```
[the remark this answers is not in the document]
timurgilfanov: Fixed. Removed the inline `//` block and moved the DI-cycle rationale into a `/** ... */` KDoc above `provideTokenRefreshUseCase`.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/TokenRefreshUseCaseImpl.kt
```diff
@@ -1,32 +1,48 @@
...
+        val refreshTokenBeforeRefresh = (stateBeforeRefresh as? AuthState.Authenticated)
+            ?.session?.tokens?.refreshToken
+
+        return authRepository.refreshToken().fold(
+            onSuccess = { tokens ->
+                if (scopeKeyBeforeRefresh != null &&
+                    tokens.refreshToken != refreshTokenBeforeRefresh
+                ) {
+                    deleteUserDataLoggingFailure(scopeKeyBeforeRefresh, "token rotation")
+                }
```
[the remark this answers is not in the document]
timurgilfanov: No action. If `restoreAuthState()` throws (now guarded by the `try/finally` fix on the `init` block), auth state stays `Unauthenticated` and the scope key is unknown — cleanup cannot run regardless. Deriving the scope key from the outgoing refresh token instead would require exposing which token `authRepository.refreshToken()` used, a significant API change for a negligible practical impact.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/LogoutUseCaseImpl.kt
```diff
@@ -1,20 +1,45 @@
...
+        } else {
+            null
+        }
+        val result = authRepository.logout().mapError { error ->
             logger.e(TAG, "Repository logout failed: $error")
             error.toUseCaseError()
         }
+        if (userKey != null) {
+            val sessionLocallyCleared = result !is ResultWithError.Failure ||
+                result.error !is LogoutError.LocalOperationFailed
```
timurgilfanov: This looks like an assumption on the order of local and remote operations. If remote operation failed, local never executed, we delete user data from settings, but i'm not sure that our auth status changed to unauthenticated.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImpl.kt
```diff
@@ -467,6 +469,30 @@ class SettingsRepositoryImpl @Inject constructor(
             },
         )
 
+    override suspend fun deleteUserData(
+        userKey: UserScopeKey,
+    ): ResultWithError<Unit, DeleteUserDataRepositoryError> {
+        runCatching {
```
timurgilfanov, line 475: Nice to have catching in the data source layer and clear contract on boundary between repo and data source.

core/domain/src/testFixtures/kotlin/timur/gilfanov/messenger/domain/usecase/settings/SettingsRepositoryFake.kt
```diff
@@ -58,4 +64,16 @@ class SettingsRepositoryFake(
...
     ): ResultWithError<Unit, SyncAllSettingsRepositoryError> =
         error("syncAllPendingSettings not configured for this test")
+
+    override suspend fun deleteUserData(
+        userKey: UserScopeKey,
+    ): ResultWithError<Unit, DeleteUserDataRepositoryError> {
+        lastDeleteUserDataKey = userKey
+        if (deleteUserDataResult is ResultWithError.Success) {
+            settingsFlow.update {
+                ResultWithError.Failure(GetSettingsRepositoryError.SettingsResetToDefaults)
```
timurgilfanov, line 74: Why we update settings flow on user data delete?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/TokenRefreshUseCase.kt
```diff
@@ -6,10 +6,14 @@ import timur.gilfanov.messenger.domain.entity.auth.AuthTokens
 /**
  * Refreshes the current access token using the stored refresh token.
  *
- * On terminal errors ([TokenRefreshError.SessionExpired]), a logout is performed as a side effect
- * to transition the auth state to [timur.gilfanov.messenger.domain.entity.auth.AuthState.Unauthenticated].
- * The logout result is ignored; callers should treat [TokenRefreshError.SessionExpired] as the
- * authoritative signal that the session has ended.
+ * On terminal refresh failures (token expired or revoked), a logout is performed to transition
+ * the auth state to [timur.gilfanov.messenger.domain.entity.auth.AuthState.Unauthenticated] and
+ * delete user-scoped settings data. If the local session clearing fails, [TokenRefreshError.LocalOperationFailed]
```
timurgilfanov, line 11: Is it right to clean settings where we write unauthenticated state to `AuthRepositoty`? May be it will be cleaner to listen to auth state and when in transit to unauthenticated clean settings? There are more than one writer and one behavior that attached to this event.

timurgilfanov:
Addressing the pending review comments:

---

**`LogoutUseCaseImpl.kt` — assumption on local/remote ordering**

Valid concern. The `sessionLocallyCleared` condition was using the error type to indirectly infer whether the local session was cleared — coupling the use case to repository ordering internals that aren't visible from the contract.

Fixed: replaced the block with a direct auth state observation:
```kotlin
if (userKey != null && authRepository.authState.first() is AuthState.Unauthenticated) {
```
Intent is now explicit: run cleanup only when the user is actually unauthenticated. `AuthRepositoryFake.logout()` already models the state transition correctly, so all existing tests pass without modification.

---

**`SettingsRepositoryImpl.kt` — "nice to have catching at the boundary"**

Agreed. The `runCatching` at this boundary is a workaround for `cancelUserScopedJobs` returning `Unit` instead of a typed result. Typed error signaling from the scheduler would remove the workaround and make the contract explicit at the interface level. Tracked in #344.

---

**`SettingsRepositoryFake.kt` — why update settings flow on delete?**

The fake models the real repository contract: after a successful `deleteUserData`, the user's settings rows are gone from the DB. Emitting `SettingsResetToDefaults` signals this to any observer still holding the flow. Without this, a test that calls `observeSettings` after a successful delete would still see the pre-delete `Settings` value and produce a phantom pass.

---

**`TokenRefreshUseCase.kt` — reactive cleanup listener**

Valid architectural concern. A reactive observer on `authState` would give cleanup-for-free to any future path that writes `Unauthenticated`, removing the implicit convention that each use case must remember to call `deleteUserData`.

Out of scope for this PR — it requires a new component, removing cleanup from both `LogoutUseCaseImpl` and `TokenRefreshUseCaseImpl`, handling the initial state edge case, and updating all related tests. Tracked in #345.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/LogoutUseCaseImpl.kt
timurgilfanov: Valid concern. The `sessionLocallyCleared` condition was using the error type to indirectly infer whether the local session was cleared — coupling the use case to repository ordering internals that are not part of the interface contract.

Fixed: replaced the block with a direct auth state observation:
```kotlin
if (userKey != null && authRepository.authState.first() is AuthState.Unauthenticated) {
```
Intent is now explicit: run cleanup only when the user is actually unauthenticated. `AuthRepositoryFake.logout()` already models the state transition correctly, so all existing tests pass without modification.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImpl.kt
timurgilfanov: Agreed. The `runCatching` at this boundary is a workaround for `cancelUserScopedJobs` returning `Unit` instead of a typed result. Typed error signaling from the scheduler would remove the workaround and make the contract explicit at the interface level. Tracked in #344.

core/domain/src/testFixtures/kotlin/timur/gilfanov/messenger/domain/usecase/settings/SettingsRepositoryFake.kt
timurgilfanov: The fake models the real repository contract: after a successful `deleteUserData`, the user's settings rows are gone from the DB. Emitting `SettingsResetToDefaults` signals this to any observer still holding the flow. Without this, a test that calls `observeSettings` after a successful delete would still see the pre-delete `Settings` value and produce a phantom pass.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/TokenRefreshUseCase.kt
timurgilfanov: Valid architectural concern. A reactive observer on `authState` would give cleanup-for-free to any future path that writes `Unauthenticated`, removing the implicit convention that each use case must remember to call `deleteUserData`.

Out of scope for this PR — it requires a new component, removing cleanup from both `LogoutUseCaseImpl` and `TokenRefreshUseCaseImpl`, handling the initial-state edge case, and updating all related tests. Tracked in #345.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncScheduler.kt
timurgilfanov: Out of scope for this PR. The `runCatching` in `SettingsRepositoryImpl.deleteUserData` already handles the exception case and preserves the best-effort contract. Changing the interface to `ResultWithError<Unit, CancelJobsError>` is a valid data-layer consistency improvement but requires adding `CancelJobsError`, making the function `suspend`, awaiting the WorkManager `Operation`, updating the repository, and updating test doubles. Tracked in #344.

PR #349, opened 2026-04-18
Fix SyncSettingWorker race with deleteUserData cleanup

### LINKED ISSUE #343: [Bug] SyncSettingWorker may recreate deleted settings rows after deleteUserData cleanup
## Bug description

In `SettingsRepositoryImpl.deleteUserData`, `cancelUserScopedJobs()` calls `workManager.cancelAllWorkByTag(userKey.key)` which enqueues cancellation asynchronously but does not wait for it. A currently-running `SyncSettingWorker` can still complete its write to the local database after `deleteAllForUser()` removes the rows, recreating stale settings data for the deleted user scope.

This is a theoretical race: cancellation is enqueued first, then local rows are deleted. If a worker happens to be mid-execution, it may write to the database after deletion. Awaiting the `Operation` returned by `cancelAllWorkByTag` would reduce the window (ensuring the cancellation is at least registered in WorkManager's internal DB before deletion) but would not eliminate it, because WorkManager cancellation is cooperative — the worker must check `isStopped` before each write to stop immediately.

## Steps to reproduce

1. Schedule a `SyncSettingWorker` for the current user.
2. While the worker is executing its DB write, trigger logout.
3. `deleteUserData` cancels jobs then deletes local rows.
4. Worker completes its write after deletion, recreating the row.

This race is unlikely in practice due to timing, but is theoretically possible on slow I/O paths.

## Expected behavior

After `deleteUserData` completes, no settings rows exist for the outgoing `UserScopeKey`.

## Actual behavior

If a `SyncSettingWorker` was mid-execution at cancellation time, it may write a setting row after deletion, leaving stale data.

## Build flavor

All

## Affected layer

Data

## Suggested priority

low

## Investigation

Decide whether to:
1. Await the `Operation` from `cancelAllWorkByTag` before calling `deleteAllForUser` (reduces window, requires `cancelUserScopedJobs` to become `suspend`).
2. Guard `SyncSettingWorker`'s DB write with `isStopped` checks so it exits early on cancellation.
3. Accept the race as negligible given the "best-effort cleanup" design contract of `deleteUserData`.

Related: PR #342 which introduced `deleteUserData`.

timurgilfanov:
Closes #343

## Summary
Eliminates the race where a running `SyncSettingWorker` can recreate user-scoped settings rows after logout cleanup deletes them.

The fix applies both sides of the mitigation from issue #343: awaiting WorkManager cancellation reduces the race window, and explicit `isStopped` checks in `SyncSettingWorker` prevent a cancelled worker from reporting or preserving a successful sync after cancellation. This keeps the cleanup contract aligned with the existing best-effort deletion flow without adding stronger persistence guarantees than WorkManager can provide.

## Changes
- make `SettingsSyncScheduler.cancelUserScopedJobs` suspending so `SettingsRepository.deleteUserData` can await WorkManager cancellation registration before deleting local rows
- make `SyncSettingWorker` cooperate with cancellation by aborting before work starts and discarding success if it is stopped during sync
- add repository and worker tests that cover cancellation ordering and stopped-worker behavior

## Testing
Describe how the changes were tested.
- [ ] Manually tested
- [x] Edge cases considered

Testing commands used:
- `./gradlew :app:testMockDebugUnitTest`
- `./gradlew :app:compileMockDebugAndroidTestKotlin`

## Acceptance criteria
- [ ] After `SettingsRepository.deleteUserData` completes, no settings rows exist for the outgoing `UserScopeKey`.
- [ ] A `SyncSettingWorker` cancelled does not recreate deleted settings rows after cleanup.

## Follow-ups
- [ ] None currently identified.


app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncScheduler.kt
```diff
@@ -27,6 +27,9 @@ interface SettingsSyncScheduler {
      * Jobs are identified by the tag equal to [UserScopeKey.key] that is added when scheduling
      * via [scheduleSettingSync]. This should be called on logout so that no stale work runs
      * for the outgoing user.
+     *
+     * Suspends until WorkManager has registered the cancellation so that callers can safely
+     * perform follow-up cleanup (e.g. deleting user-scoped rows) without racing in-flight workers.
```
timurgilfanov: Must be removed because this is detail of implementation, not part of the contract.
```suggestion
```

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncSchedulerImpl.kt
```diff
@@ -55,8 +56,8 @@ class SettingsSyncSchedulerImpl @Inject constructor(private val workManager: Wor
         )
     }
 
-    override fun cancelUserScopedJobs(userKey: UserScopeKey) {
-        workManager.cancelAllWorkByTag(userKey.key)
+    override suspend fun cancelUserScopedJobs(userKey: UserScopeKey) {
+        workManager.cancelAllWorkByTag(userKey.key).result.await()
```
timurgilfanov, line 60: Is it worth to add dependency for this one usage?

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncScheduler.kt
timurgilfanov: Removed in f433e1be.

app/src/main/java/timur/gilfanov/messenger/data/repository/SettingsSyncSchedulerImpl.kt
timurgilfanov: Keeping the dep. `kotlinx-coroutines-guava` is the idiomatic bridge from `ListenableFuture` to `suspend` — inlining `suspendCancellableCoroutine` here would trade one line for a ~10-line local wrapper of equivalent behavior. Small, vetted, ecosystem-standard; worth it even for a single call site.

PR #352, opened 2026-04-23
feat: validate that password is not equal to email

timurgilfanov:
Closes #328

## Summary
Adds a cross-field credentials validation rule that rejects passwords matching the entered email address, including the user-facing error string and localized German resource.

The equality check remains inside `CredentialsValidatorImpl` so credentials validation stays centralized in the feature domain validator. The check is performed after the existing password rules, preserving the current validation ordering while adding the cross-field rule requested by the issue.

## Changes
- Added `PasswordValidationError.PasswordEqualToEmail` and mapped it to auth display strings.
- Updated `CredentialsValidatorImpl` to run the password-equals-email check after existing password validation succeeds.
- Added and updated auth validator and real-time ViewModel tests for the new validation path.

## Testing
Validation recorded in the completed plan:
- [ ] Manually tested
- [x] Edge cases considered

Commands recorded as run:
- `./gradlew :feature:auth:testDebugUnitTest --tests "timur.gilfanov.messenger.auth.domain.validation.CredentialsValidatorImplTest"`

### DIFF
--- a/feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
+++ b/feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
@@ -69,6 +69,7 @@ class LoginViewModel @Inject constructor(
                 Email(email),
             ) as? ResultWithError.Failure
             )?.error
+        // Password error depends on email via `PasswordError.PasswordEqualToEmail`
         val passwordError = when {
             currentPassword.isEmpty() -> _state.value.passwordError
             credError is CredentialsValidationError.Password -> credError.reason
@@ -96,6 +97,7 @@ class LoginViewModel @Inject constructor(
         )
         val isCredentialsValid = credentialsResult is ResultWithError.Success
         val credError = (credentialsResult as? ResultWithError.Failure)?.error
+        // Email check have a priority in credentials validation in `CredentialsValidatorImpl`
         val passwordError = when (credError) {
             is CredentialsValidationError.Password -> credError.reason
             else -> (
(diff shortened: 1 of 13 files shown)


feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
```diff
@@ -58,30 +59,51 @@ class LoginViewModel @Inject constructor(
...
+        )
+        val isCredentialsValid = credentialsResult is ResultWithError.Success
+        val credError = (credentialsResult as? ResultWithError.Failure)?.error
+        val passwordError = when (credError) {
+            is CredentialsValidationError.Password -> credError.reason
+            else -> (
+                credentialsValidator.validate(
+                    Password(password),
+                ) as? ResultWithError.Failure
+                )?.error
```
timurgilfanov, line 107: Why we need to validate password here?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
```diff
@@ -58,30 +59,51 @@ class LoginViewModel @Inject constructor(
...
+        val passwordError = when {
+            currentPassword.isEmpty() -> _state.value.passwordError
+            credError is CredentialsValidationError.Password -> credError.reason
+            credentialsResult is ResultWithError.Success -> null
+            else -> (
+                credentialsValidator.validate(
+                    Password(currentPassword),
+                ) as? ResultWithError.Failure
+                )?.error
+        }
```
timurgilfanov, line 82: Nice to add an explanation comment for this complex expression.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
timurgilfanov: To set password error in case there is active email error because email check have a priority in credentials validation in `CredentialValidatorImpl`.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/ui/screen/login/LoginViewModel.kt
timurgilfanov: ok

PR #353, opened 2026-04-23
Fix flaky AuthInterceptor concurrent refresh test

### LINKED ISSUE #350: [Bug] Flacky concurency test of `AuthInterceptor`
### Bug description

`AuthInterceptorTest.when concurrent 401 responses then tokenRefreshUseCase is invoked only once`

java.lang.AssertionError: expected:<1> but was:<2>

### Steps to reproduce

Run unit components tests on CI. 

### Expected behavior

Test always passed.

### Actual behavior

Sometimes test failed on CI.

### Build flavor

mock

### Android API level

_No response_

### Device / Emulator

_No response_

### Affected layer

None

### Suggested priority

None

### Logs, screenshots, or stack traces

```shell

```

timurgilfanov:
Closes #350

## Summary
Fixes the flaky concurrent `AuthInterceptor` 401 test by replacing fragile scheduler coordination with deterministic test-scheduler draining before the blocked refresh is released.

Production `AuthInterceptor` code was left unchanged. The issue was isolated to test scheduling: `runCurrent()` only processes work already queued at that moment, so some concurrent requests could arrive after the refresh was released and observe no active refresh. `advanceUntilIdle()` drains pending test work until all launched requests are suspended on the shared refresh.

## Changes
- Replaced the `yield()` loop and `runCurrent()` synchronization in the concurrent 401 test with `advanceUntilIdle()`.
- Collected concurrent responses with `async`/`awaitAll` and asserted they all return `Unauthorized` when refresh fails.
- Added a regression test proving a later 401 after a completed refresh starts a new refresh.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

Validation recorded in the completed plan:
- `./gradlew :feature:auth:testDebugUnitTest`

## Acceptance criteria
- [ ] The concurrent 401 `AuthInterceptor` unit test passes consistently.

### DIFF
--- a/feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/AuthInterceptorTest.kt
+++ b/feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/AuthInterceptorTest.kt
@@ -5,13 +5,12 @@ import io.ktor.client.engine.mock.MockEngine
 import io.ktor.client.engine.mock.respond
 import io.ktor.client.request.get
 import io.ktor.http.HttpStatusCode
-import java.util.concurrent.atomic.AtomicInteger
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.joinAll
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.advanceUntilIdle
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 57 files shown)


feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/AuthInterceptorTest.kt
```diff
@@ -231,45 +221,90 @@ class AuthInterceptorTest {
...
-        }
-        runCurrent()
+        advanceUntilIdle()
         releaseRefresh.complete(Unit)
-        jobs.joinAll()
+        val responses = jobs.awaitAll()
 
-        assertEquals(1, refreshInvocationCount.get())
+        assertEquals(1, refreshInvocationCount)
+        responses.forEach { assertEquals(HttpStatusCode.Unauthorized, it.status) }
```
timurgilfanov, line 309: This check is out of the scope of the test and must be removed.
```suggestion
```

timurgilfanov: This line checks the other half of the interceptor contract: when that shared refresh fails, every waiting caller gets the original `401 Unauthorized` back.

PR #354, opened 2026-04-23
Stabilize auth rotation instrumentation tests

### LINKED ISSUE #340: [Bug] LoginFeatureTest loginScreen_handlesRotation is flaky and sometimes times out in Firebase Test Lab CI
### Bug description
`feature/auth/src/androidTest/java/timur/gilfanov/messenger/feature/auth/LoginFeatureTest.kt` test `loginScreen_handlesRotation` is flaky in GitHub Actions CI and sometimes fails with a timeout during the Firebase Test Lab instrumentation run.

The test does not fail every time. It passes on some runs and times out on others, which indicates a synchronization issue around configuration change handling rather than a deterministic functional failure.

Firebase Test Lab log analysis shows that rotation itself succeeds: the activity is recreated and reaches `RESUMED` again after the orientation change. The test then hangs because `waitForIdle()` never returns, so the test runner never reaches the final assertions or test completion.

This points to a post-rotation idling problem, likely involving Compose/Espresso synchronization with IME and window redraw activity, not a crash in login screen business logic.

### Steps to reproduce
1. Run the `feature-tests` GitHub Actions workflow for a PR that includes `:feature:auth` instrumentation tests.
2. Let the `Run feature:auth Feature Instrumentation Tests on Emulators` step execute on Firebase Test Lab (`Pixel2.arm`, API 30).
3. Repeat the workflow across multiple runs.
4. Observe that `LoginFeatureTest.loginScreen_handlesRotation` sometimes passes and sometimes times out.

### Expected behavior
The rotation test should wait deterministically for the recreated login screen to be rendered after orientation change and pass reliably in CI on every run.

### Actual behavior
The test is flaky in CI and intermittently times out during or after rotation.

### Build flavor
mock

### Android API level
30

### Device / Emulator
Firebase Test Lab `Pixel2.arm` emulator

### Affected layer
UI

### Suggested priority
medium

### Logs, screenshots, or stack traces
Relevant code path:

`LoginFeatureTest.kt`
```kotlin
activity.requestedOrientation = SCREEN_ORIENTATION_LANDSCAPE
waitForIdle()
onNodeWithTag("login_email_field").assertExists()
```

Confirmed from Firebase Test Lab log:
- `loginScreen_handlesRotation` starts normally
- the rotated `LoginScreenTestActivity` is recreated and returns to `RESUMED`
- there is no crash, ANR, or assertion failure in the app process
- Compose registers `androidx.compose.ui.test.EspressoLink`, but unlike passing tests there is no matching idling-resource unregister and no `finished: loginScreen_handlesRotation(...)`
- the log around the hang contains repeated window/IME churn after rotation, including `Orientation start waiting for draw`, `IInputConnectionWrapper ... on inactive InputConnection`, and `Failed looking up window session`

Likely cause:
- `waitForIdle()` is used as the only post-rotation synchronization point
- rotation triggers async activity recreation plus IME/window reattachment work
- on slower Firebase Test Lab emulators, Compose/Espresso sometimes never observes the app as idle again, so the test hangs until CI times out

Potential fix:
- Replace post-rotation `waitForIdle()` with an explicit wait for a post-recreation node, for example `waitUntilExactlyOneExists(hasTestTag("login_email_field"), timeoutMillis = SCREEN_LOAD_TIMEOUT_MILLIS)`
- Audit similar auth rotation tests that use the same pattern

timurgilfanov:
Closes #340

## Summary
Stabilizes the auth rotation instrumentation tests that were flaky in Firebase Test Lab by avoiding post-rotation `waitForIdle()` synchronization and waiting for concrete UI nodes instead.

Post-rotation tests now use a two-phase wait: first observe the old screen leaving composition, then wait for the recreated screen's expected node. This avoids relying on global idleness when Firebase Test Lab keeps Compose/Espresso busy with IME and window redraw work.

## Changes
- Replaces post-rotation synchronization in login and signup feature tests with explicit screen disappearance and node reappearance waits.
- Adds timeouts to several auth error/dialog waits so failures are bounded and easier to diagnose.
- Documents the post-rotation synchronization convention in `CLAUDE.md`.

## Testing
- [x] Tests added or updated
- [ ] Manually tested
- [x] Edge cases considered

## Acceptance criteria
- [x] `LoginFeatureTest` and `SignupFeatureTest` passed constantly on CI.


CLAUDE.md
```diff
@@ -47,6 +47,7 @@ Full strategy in `docs/Testing Strategy.md`.
...
 - **`assertTrue` over `assert`**: Use `assertTrue` from `kotlin.test` for boolean assertions in tests. Kotlin's built-in `assert()` is a JVM assertion that can be silently disabled at runtime with `-da`.
+- **Post-rotation synchronization**: After `activity.requestedOrientation = SCREEN_ORIENTATION_*`, never use `waitForIdle()` (hangs on Firebase Test Lab due to IME and window-redraw idling) and never call `waitUntilExactlyOneExists` on a node that already exists pre-rotation (satisfies immediately on the stale composition). Capture `val oldActivity = activity` before the rotation, then call `waitUntil(timeoutMillis = …) { oldActivity.isDestroyed }` to observe Activity destruction (do NOT use `waitUntilDoesNotExist(hasTestTag("…_screen"))` — on fast emulators the new Activity's node reappears before the first poll, making the condition unsatisfiable), then `waitUntilExactlyOneExists(hasTestTag("…"), timeoutMillis = …)` to confirm the post-rotation composition is ready.
```
timurgilfanov, line 50: Not clear what this means: "(do NOT use `waitUntilDoesNotExist(hasTestTag("…_screen"))` — on fast emulators the new Activity's node reappears before the first poll, making the condition unsatisfiable)".

timurgilfanov: `waitUntilDoesNotExist(hasTestTag("…_screen"))` sounds like it waits for the old screen to disappear after rotation, but Compose test matchers do not know whether the matching node belongs to the old Activity or the new Activity. They only ask: “does any node with this tag currently exist?”

On a fast emulator, this can happen:

1. Old Activity has login_screen.
2. Test triggers rotation.
3. Old Activity is destroyed.
4. New Activity is created very quickly.
5. New composition also contains login_screen.
6. The first poll of `waitUntilDoesNotExist(hasTestTag("login_screen"))` happens after step 5.

From the test’s point of view, login_screen still exists, even though it is now the new Activity’s node. Since the new screen is supposed to exist after rotation, the condition “no node with this tag exists” may never become true. That makes the wait time out.

The safer pattern is:
```
val oldActivity = activity

activity.requestedOrientation = SCREEN_ORIENTATION_LANDSCAPE

composeRule.waitUntil(timeoutMillis = ...) {
    oldActivity.isDestroyed
}

composeRule.waitUntilExactlyOneExists(
    hasTestTag("login_screen"),
    timeoutMillis = ...
)
```
This waits for the actual old Activity instance to be destroyed, then waits for the new composition to appear. It avoids confusing the old screen node with the new screen node just because they share the same test tag.

PR #362, opened 2026-04-30
Centralize auth cleanup observation

### LINKED ISSUE #345: [Refactor] Observe auth state transitions to trigger settings cleanup reactively
## Current state

`LogoutUseCaseImpl` and `TokenRefreshUseCaseImpl` each call `settingsRepository.deleteUserData()` imperatively at the point where they transition auth state to `Unauthenticated`. Any future code path that writes `Unauthenticated` must remember to trigger the same cleanup, creating an implicit convention rather than an enforced contract.

## Desired state

A single reactive component observes `authRepository.authState` as a `Flow`, detects `Authenticated → Unauthenticated` transitions, captures the outgoing `UserScopeKey`, and calls `settingsRepository.deleteUserData(outgoingKey)`. Cleanup calls are removed from `LogoutUseCaseImpl` and `TokenRefreshUseCaseImpl`.

## Motivation

Single responsibility for settings cleanup. New paths to `Unauthenticated` (e.g., forced logout, admin revocation) receive cleanup automatically without any changes to individual use cases.

## Alternatives considered

Keep the current imperative approach — simpler, explicit, all paths currently tested. The risk is that future writers to `Unauthenticated` miss the cleanup step.

## Affected files or modules

- `feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/LogoutUseCaseImpl.kt`
- `feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/domain/usecase/TokenRefreshUseCaseImpl.kt`
- New observer component (location TBD — likely `:feature:auth` domain or data layer)
- Related tests

timurgilfanov:
Closes #345

## Summary
Centralizes user-scoped settings cleanup behind a reactive auth cleanup observer. This removes the implicit requirement for each auth use case that writes `Unauthenticated` to remember to delete user data itself.

## Changes
- Add `AuthCleanupObserver` to observe `AuthRepository.authState` transitions.
- Clean up the outgoing authenticated user scope on `Authenticated -> Unauthenticated` transitions.
- Clean up the previous user scope when an `Authenticated -> Authenticated` transition changes `UserScopeKey`, such as refresh-token rotation.
- Initialize the observer from auth repository setup so cleanup is owned centrally.
- Remove direct settings cleanup calls from `LogoutUseCaseImpl` and `TokenRefreshUseCaseImpl`.
- Serialize logout local-session clearing with refreshed-token persistence so a concurrent refresh cannot restore credentials after logout.
- Add and update auth tests covering observer behavior, repository transitions, logout, token refresh, signup, and the refresh/logout race.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

## Screenshots / Demo
Not applicable; this refactor does not change UI.

## Breaking Changes
None expected.

## Follow-ups
- [x] #363


feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/source/local/LocalAuthDataSource.kt
```diff
@@ -19,4 +20,36 @@ interface LocalAuthDataSource {
...
+     * The comparison and removal happen atomically in a single transaction.
+     */
+    suspend fun clearPendingCleanupKeyIfMatches(
+        key: UserScopeKey,
+    ): ResultWithError<Unit, LocalAuthDataSourceError>
+
+    /**
+     * Returns the pending cleanup key written by a previous session, or null if none exists.
+     */
+    suspend fun getPendingCleanupKey(): ResultWithError<UserScopeKey?, LocalAuthDataSourceError>
```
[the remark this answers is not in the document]
timurgilfanov: - The single-slot design is intentional. Both AuthCleanupObserver.handleScopeChange (line 124) and
 AuthRepositoryImpl.writeTokenRotationCleanupMarkerIfNeeded (line 321) explicitly log "Another cleanup is
  already pending; running cleanup … best-effort." and continue.
 - The KDoc on setPendingCleanupKeyIfAbsent (line 33–41) documents the single-slot semantics as the
 contract.
 - No bug, test failure, or open issue indicates this trade-off is causing problems. The recovery is
 "best-effort": at most one outstanding key survives process death; subsequent scope-change cleanups run
 in-memory only.
 - A move to multi-entry would touch the interface, LocalAuthDataSourceImpl (DataStore-backed), both
 LocalAuthDataSourceFake versions, AuthRepositoryImpl, AuthCleanupObserver, and ~20 tests across
 AuthCleanupObserverTest and AuthRepositoryImplTest. It's an architectural redesign, not a code-review
 fix.

 Verdict: Do not apply. Out of scope for a review-fix pass. If we want to pursue stronger durability
 guarantees for cleanup, raise a separate issue + ADR.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -41,6 +44,7 @@ import timur.gilfanov.messenger.util.Logger
 class AuthRepositoryImpl @Inject constructor(
     private val remoteDataSource: RemoteAuthDataSource,
     private val localDataSource: LocalAuthDataSource,
+    private val cleanupObserver: Lazy<AuthCleanupObserver>,
```
timurgilfanov: Nice to have cleaner approach that removes `AuthCleanupObserver` from `AuthRepositoryImpl`. Observation could be started from app initializer.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -52,7 +56,11 @@ class AuthRepositoryImpl @Inject constructor(
     private val _authState = MutableStateFlow<AuthState?>(null)
     override val authState: Flow<AuthState> = _authState.filterNotNull()
 
+    // Prevents refresh from saving new tokens while logout is clearing the local session.
+    private val sessionMutex = Mutex()
```
timurgilfanov, line 62: Good enough for now. This protection is partial and execution order depends on coroutines scheduler. To remove this ambiguity we can serialize events of logout and token refresh and process them with one coroutine.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserver.kt
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Reactively cleans up user-scoped data when the user logs out or session expires.
+ *
+ * Observes [AuthRepository.authState] and triggers [SettingsRepository.deleteUserData]
+ * on Authenticated -> Unauthenticated transitions or when the user scope key changes
+ * (e.g. during token rotation if it affects the key).
+ */
+@Singleton
+class AuthCleanupObserver @Inject constructor(
```
timurgilfanov: Must inherit interface to easy replacement with test doubles.

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImplTest.kt
```diff
@@ -222,33 +229,123 @@ class AuthRepositoryImplTest {
...
+        assertIs<RefreshRepositoryError.SessionRevoked>(refreshFailure.error)
+        assertNull(storage.accessToken)
+        assertNull(storage.refreshToken)
+        repo.authState.test {
+            assertIs<AuthState.Unauthenticated>(awaitItem())
+        }
+    }
+
+    @Test
+    fun `refresh during logout-then-login returns SessionRevoked without overwriting tokens`() =
```
timurgilfanov, line 319: Nice to add test for returning SessionRevoked when refresh token changed during `refreshToken()` call if this test doesn't cover that case.

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/AuthInterceptorTest.kt
```diff
@@ -276,7 +277,9 @@ class AuthInterceptorTest {
      */
     @OptIn(ExperimentalCoroutinesApi::class)
     @Test
-    fun `when concurrent 401 responses then tokenRefreshUseCase is invoked only once`() = runTest {
+    fun `when concurrent 401 responses then tokenRefreshUseCase is invoked only once`() = runTest(
+        UnconfinedTestDispatcher(),
```
timurgilfanov: Do we need this dispatcher?

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/TokenRefreshUseCaseTest.kt
```diff
@@ -129,121 +119,74 @@ class TokenRefreshUseCaseTest {
...
-        assertIs<TokenRefreshError.LocalOperationFailed>(result.error)
-        assertIs<AuthState.Authenticated>(authRepositoryFake.currentAuthState)
-        kotlin.test.assertNull(settingsRepository.lastDeleteUserDataKey)
+        assertIs<TokenRefreshError.SessionExpired>(result.error)
+        assertIs<AuthState.Unauthenticated>(authRepositoryFake.currentAuthState)
     }
 
     @Test
-    fun `when SessionRevoked and local logout fails then returns LocalOperationFailed`() = runTest {
+    fun `when SessionRevoked and remote logout fails then returns SessionExpired`() = runTest {
```
timurgilfanov, line 176: Is this correct behaviour?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
timurgilfanov: Done. `AuthCleanupObserver.start()` now runs from `MessengerApplication.onCreate()`; `AuthRepositoryImpl` no longer takes the observer as a constructor parameter. Also lifted the `@ApplicationScope CoroutineScope` provider out of `AuthDataModule` into a dedicated `AuthApplicationScopeModule` so existing tests that `@UninstallModules(AuthDataModule::class)` keep building.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
timurgilfanov: Closed the named race in this PR (`saveRefreshedTokens` identity check + extended `sessionMutex` coverage on login/signup writes). The residual ordering ambiguity (which concurrent event "wins" depends on the dispatcher) is tracked in #364; that issue intentionally lists several possible solutions (single-coroutine actor, cancel-on-newer-event, sequence numbers, event-sourcing) without prescribing one.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserver.kt
timurgilfanov: Done. `AuthCleanupObserver` is now an interface with `AuthCleanupObserverImpl`; bound via `AuthCleanupObserverModule` (mirrors `SettingsSyncSchedulerModule`). Tests can substitute a no-op double.

timurgilfanov:
@coderabbitai re: outside-diff review on `AuthRepositoryImpl.kt` lines 222-284 (extend the session lock beyond logout/refresh) — closed in commit d0ee9dc7. `saveRefreshedTokens` now compares the refresh token used against `currentState.session.tokens.refreshToken` under the mutex and returns `SessionRevoked` if they differ; `loginWithCredentials`, `loginWithGoogle`, `signup`, and `signupWithGoogle` now wrap `saveSession` + `setState` in `sessionMutex.withLock`. Added regression test `refresh during logout-then-login returns SessionRevoked without overwriting tokens`. The residual ordering ambiguity (concurrent events resolve in dispatcher acquire-order, not initiation order) is tracked separately in #364.

timurgilfanov:
@coderabbitai re: outside-diff review on `AuthRepositoryImpl.kt` lines 223-255 (re-check active session before clearing in `logout()`) — closed in commit 061a1a01. `logout()` now re-reads the access token under `sessionMutex`; if the active session was replaced during the remote logout call (a concurrent login wrote a new session), the local clear and state-write are skipped and the result is `Success(Unit)` (or the remote error if there was one). Added regression test `logout during concurrent login does not wipe the new session`. Residual cross-event ordering is still tracked in #364.

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImplTest.kt
timurgilfanov: Already covered by the existing test `refresh during logout-then-login returns SessionRevoked without overwriting tokens` (commit d0ee9dc7, lines ~287-330). The scenario it exercises is exactly "refresh token changed during `refreshToken()` call": refresh starts with refresh-token A, in flight; logout completes and a login as user B writes a new session with refresh-token B; refresh resumes; the identity check in `saveRefreshedTokens` (`currentSession.tokens.refreshToken != refreshTokenUsed`) catches the mismatch and returns `SessionRevoked` without overwriting B's tokens.

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/AuthInterceptorTest.kt
timurgilfanov: No — removed in commit 061a1a01. The KDoc only justifies `mockEngine.config.dispatcher = Dispatchers.Unconfined` (which is still required so MockEngine doesn't run handlers on real IO threads that `advanceUntilIdle` cannot drain). The `UnconfinedTestDispatcher()` passed to `runTest` was redundant; verified empirically the test still passes deterministically with the default `StandardTestDispatcher`.

feature/auth/src/test/kotlin/timur/gilfanov/messenger/auth/TokenRefreshUseCaseTest.kt
timurgilfanov: The test reflects the current implementation correctly, but the underlying flow has three implicit design choices worth re-examining (collapsing TokenExpired/SessionRevoked into SessionExpired, treating transient logout failures as terminal, and hiding a logout side-effect inside `refreshToken()`). Those design questions are tracked separately in #366 so they don't get lost; this PR doesn't change the semantics.

timurgilfanov:
@coderabbitai re: nitpicks on review id=4210572376 — both addressed in commit 061a1a01.

- `AuthRepositoryImpl.kt:57-60` (sessionMutex inline comment): converted the multi-line `//` rationale to a KDoc on the property, matching the no-inline-comments style.
- `AuthCleanupObserverImpl.kt:67-71` (UserScopeKey leakage in log): `deleteUserData` no longer interpolates the `UserScopeKey` into the failure log; only the underlying `result.error` is logged. UserScopeKey is sha256(refreshToken) — a stable per-user identifier, so dropping it avoids cross-session correlation by anyone with log access.

timurgilfanov:
@coderabbitai re: nitpick on review id=4214363508 (`AuthRepositoryImpl.kt` lines 306-311, inline comment on refresh-token check) — addressed in commit 82cadc30. The three-line `//` comment is removed; the race-case semantics are now documented in a KDoc on `saveRefreshedTokens` itself, matching the no-inline-comments style. The check (`currentState.session.tokens.refreshToken != refreshTokenUsed`) is unchanged.

PR #371, opened 2026-05-04
Fix settings offline fallback without syncing defaults

### LINKED ISSUE #333: [Bug] Settings offline fallback persists unsynced defaults, overwriting server preferences on reconnect
### Bug description
After the settings table is cleared (e.g. Room migration 3→4), the first app launch while offline triggers `recoverSettings()` in `SettingsRepositoryImpl`. Because the remote call fails, the code falls back to `upsertDefaultSettings()`, which persists default settings via `defaultLocalSetting()` with `localVersion=1, syncedVersion=0`.

Since `syncedVersion ≠ localVersion`, these defaults are treated as pending sync. On the next online run, `syncAllPendingSettings` picks them up via `getUnsynced()` and pushes them to the server — overwriting the user's real server-side preferences with defaults.

The `RemoteSetting.Missing` path (where the server genuinely has no value) correctly uses `localVersion=1, syncedVersion=0` because in that case we *do* want to sync the default up. The offline fallback path should not produce the same state.

### Steps to reproduce
1. Have an account with non-default settings synced to the server
2. Clear app data or trigger migration 3→4 (which drops the settings table)
3. Launch the app while offline
4. Restore network connectivity
5. Observe settings sync — server preferences are overwritten with defaults

### Expected behavior
Offline-fallback defaults are local-only placeholders. They should not be synced to the server. On next launch with network, the real server values should be fetched and applied.

### Actual behavior
Defaults created during the offline fallback are marked as unsynced (`syncedVersion=0`). They get pushed to the server on next sync, overwriting the user's real preferences.

### Build flavor
All

### Affected layer
Data

### Suggested priority
medium

### Root cause
`defaultLocalSetting()` in `LocalSettings.kt` always produces `localVersion=1, syncedVersion=0`.
`upsertDefaultSettings()` in `SettingsRepositoryImpl.kt` uses this function for the offline fallback, creating an unsynced state.

### Fix
In `upsertDefaultSettings()`, build the `LocalSetting` directly with `syncedVersion=1` (matching `localVersion`) so offline-fallback defaults are not treated as pending changes. The `defaultLocalSetting()` helper should remain unchanged since its current behaviour is correct for the `RemoteSetting.Missing` path.

### Related
PR #331

timurgilfanov:
Closes #333

## Summary
Fixes the settings offline fallback path so placeholder defaults created when remote recovery fails are not treated as pending user changes and later synced over real server preferences.

## Changes
- Preserve offline fallback settings as local placeholders instead of unsynced defaults.
- Rename the reset/default-related domain error to SettingsUnspecified to better match the behavior.
- Update settings repository, integration, use case, and fake tests around fallback and missing-remote behavior.
- Document the completed implementation plan.

## Important Design Decision
Offline fallback defaults and genuinely missing remote settings are now distinct states. Missing remote settings can still be synced as intended, while fallback placeholders remain local until real remote settings can be recovered.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

## Screenshots / Demo
Not applicable; this PR does not affect UI.

## Acceptance criteria
- [x] Offline-fallback defaults are local-only placeholders.
- [x] Offline-fallback defaults are not synced to the server.
- [x] On the next launch with network, real server values are fetched and applied.
- [x] The existing RemoteSetting.Missing path still creates syncable defaults when the server genuinely has no value.

## Breaking Changes
None.

## Follow-ups
Issues needed to be addressed but out of the scope of this PR. Mark as done when issue created.
- [ ] None

### DIFF
--- a/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
+++ b/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
@@ -80,8 +80,7 @@ class SettingsViewModel @Inject constructor(
                                     ObserveSettingsError.SettingsUnspecified -> {
                                         logger.i(
                                             TAG,
-                                            "Settings are unspecified; " +
-                                                "transient defaults emitted without persistence",
+                                            "Settings are unspecified",
                                         )
                                     }
 
--- a/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
+++ b/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
@@ -77,8 +77,12 @@ class SettingsViewModel @Inject constructor(
                                         )
                                     }
 
-                                    ObserveSettingsError.SettingsResetToDefaults -> {
-                                        logger.i(TAG, "Settings were reset to defaults")
+                                    ObserveSettingsError.SettingsUnspecified -> {
+                                        logger.i(
+                                            TAG,
+                                            "Settings are unspecified; " +
+                                                "transient defaults emitted without persistence",
+                                        )
                                     }
 
                                     is ObserveSettingsError.LocalOperationFailed -> {
--- a/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
+++ b/app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
@@ -78,10 +78,7 @@ class SettingsViewModel @Inject constructor(
                                     }
 
                                     ObserveSettingsError.SettingsUnspecified -> {
-                                        logger.i(
-                                            TAG,
-                                            "Settings are unspecified",
-                                        )
+                                        logger.i(TAG, "Settings are unspecified")
                                     }
 
                                     is ObserveSettingsError.LocalOperationFailed -> {
--- a/app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImplTest.kt
+++ b/app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImplTest.kt
@@ -143,6 +143,7 @@ class SettingsRepositoryImplTest {
         val persistedSetting = localDataSource.getSetting(testUserKey, SettingKey.UI_LANGUAGE)
         assertIs<ResultWithError.Failure<TypedLocalSetting, GetSettingError>>(persistedSetting)
         assertIs<GetSettingError.SettingNotFound>(persistedSetting.error)
+        assertTrue(trackingScheduler.scheduledSyncs.isEmpty())
     }
 
     @Test
@@ -174,7 +175,6 @@ class SettingsRepositoryImplTest {
         assertEquals(1, persistedSetting.data.setting.localVersion)
         assertEquals(0, persistedSetting.data.setting.syncedVersion)
         assertEquals(0, persistedSetting.data.setting.serverVersion)
-
         assertTrue(
             trackingScheduler.scheduledSyncs.any { (key, setting) ->
                 key == testUserKey && setting == SettingKey.UI_LANGUAGE
@@ -221,6 +221,11 @@ class SettingsRepositoryImplTest {
         assertEquals(UiLanguage.German, updatedSetting.data.setting.value)
         assertEquals(2, updatedSetting.data.setting.localVersion)
         assertEquals(1, updatedSetting.data.setting.syncedVersion)
+        assertTrue(
+            trackingScheduler.scheduledSyncs.any { (key, setting) ->
+                key == testUserKey && setting == SettingKey.UI_LANGUAGE
+            },
+        )
     }
 
     @Test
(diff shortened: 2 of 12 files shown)


core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/usecase/settings/ObserveUiLanguageError.kt
```diff
@@ -10,7 +10,7 @@ import timur.gilfanov.messenger.domain.usecase.settings.repository.GetSettingsRe
  * - [Unauthorized] - Not authenticated
  *
  * ## Logical Errors
- * - [SettingsResetToDefaults] - Settings were not found and reset to defaults
+ * - [SettingsUnspecified] - Settings are unspecified and transient defaults were emitted
```
timurgilfanov: No need to rely on repository behavior, must describe only error.
```suggestion
 * - [SettingsUnspecified] - Settings are unspecified
```

app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
```diff
@@ -77,8 +77,12 @@ class SettingsViewModel @Inject constructor(
                                         )
                                     }
 
-                                    ObserveSettingsError.SettingsResetToDefaults -> {
-                                        logger.i(TAG, "Settings were reset to defaults")
+                                    ObserveSettingsError.SettingsUnspecified -> {
+                                        logger.i(
+                                            TAG,
+                                            "Settings are unspecified; " +
+                                                "transient defaults emitted without persistence",
```
timurgilfanov: No need to rely on repository behavior, must process only error.
```suggestion
                                            "Settings are unspecified; ",
```

core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/usecase/settings/ObserveUiLanguageError.kt
timurgilfanov: Valid and in scope. Updated the KDoc to describe `SettingsUnspecified` only as the error state, and removed the repository/transient-default behavior detail from the error contract.

app/src/main/java/timur/gilfanov/messenger/ui/screen/settings/SettingsViewModel.kt
timurgilfanov: Valid and in scope. Fixed this branch to process/log only the `SettingsUnspecified` error, and swept the PR for the same coupling pattern before push: `LanguageViewModel`, `ObserveSettingsError`, `ObserveUiLanguageError`, `ObserveUiLanguageUseCase`, and `GetSettingsRepositoryError` now describe only the error state. Repository-specific transient-default wording remains only in `SettingsRepositoryImpl`, where it documents/logs the repository behavior.

app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryIntegrationTest.kt
```diff
@@ -93,28 +95,19 @@ class SettingsRepositoryIntegrationTest {
...
+                val success = awaitItem()
+                assertIs<ResultWithError.Success<Settings, *>>(success)
+                assertEquals(UiLanguage.English, success.data.uiLanguage)
 
-                    is ResultWithError.Success -> {
-                        // Success with default settings
-                        assertEquals(UiLanguage.English, result.data.uiLanguage)
-                    }
-                }
                 cancelAndIgnoreRemainingEvents()
```
timurgilfanov: Must be removed because no events are expected.
```suggestion
```

app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImplTest.kt
```diff
@@ -130,12 +133,53 @@ class SettingsRepositoryImplTest {
...
+            assertIs<GetSettingsRepositoryError.SettingsUnspecified>(result1.error)
 
             val result = awaitItem()
             assertIs<ResultWithError.Success<Settings, GetSettingsRepositoryError>>(result)
             assertEquals(defaultSettings.uiLanguage, result.data.uiLanguage)
         }
+
+        val persistedSetting = localDataSource.getSetting(testUserKey, SettingKey.UI_LANGUAGE)
+        assertIs<ResultWithError.Failure<TypedLocalSetting, GetSettingError>>(persistedSetting)
+        assertIs<GetSettingError.SettingNotFound>(persistedSetting.error)
```
timurgilfanov, line 145: Nice to remove because out of the scope of the test.
```suggestion
```

app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImplTest.kt
```diff
@@ -130,12 +133,53 @@ class SettingsRepositoryImplTest {
...
+    }
+
+    @Test
+    fun `syncAllPendingSettings is no-op after offline fallback`() = runTest {
```
timurgilfanov: Nice to specify no-op.
```suggestion
    fun `syncAllPendingSettings doesn't persist settings after offline fallback`() = runTest {
```

app/src/test/java/timur/gilfanov/messenger/data/repository/SettingsRepositoryImplTest.kt
timurgilfanov: Addressed in e46c43a6: replaced the fake’s `MutableStateFlow` with `MutableSharedFlow(replay = 1, extraBufferCapacity = 1)`, kept new collectors to one replayed latest value, and changed the delete path to ordered suspending `emit(...)` calls for `Failure(SettingsUnspecified)` then `Success(defaultSettings)`.

PR #387, opened 2026-05-07
Use message timeline as send status source of truth

### LINKED ISSUE #379: ChatViewModel represents outgoing send state
## Problem statement
Part of #213.

ChatViewModel must expose outgoing sending, sent, and failed-send states in UI state so failed messages remain visible instead of only becoming dialog errors.

## Proposed solution
Extend ChatUiState.Ready with outgoing message state representing:
- Sending
- Sent
- Failed(error)

On send failure:
- keep input stable
- do not emit ClearInputText
- keep failed outgoing message in state

On send success:
- commit successful state before emitting ClearInputText
- preserve the existing guard that only clears if input still matches submitted text

## Acceptance criteria
- [ ] Sending outgoing messages are represented in ChatUiState.
- [ ] Sent outgoing messages are represented in ChatUiState.
- [ ] Failed outgoing messages remain represented in ChatUiState.
- [ ] Failed sends keep user input stable.
- [ ] Successful sends emit ClearInputText only after success state is committed.
- [ ] Tests cover success, failure, and clear-ordering behavior.

timurgilfanov:
Closes #379

## Summary
Use the persisted chat message timeline as the canonical source of truth for outgoing send status, including sending progress and failed accepted sends.

## Changes
- Persist accepted outgoing sends locally as `DeliveryStatus.Sending(0)` before starting remote send.
- Clear composer input after local acceptance, before server acknowledgement, while preserving the existing submitted-text guard.
- Stop remote send when local persistence fails and keep composer input unchanged.
- Persist remote progress/final updates and remote failures into the same local message row.
- Preserve latest remote message fields when marking a send as failed.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

## Screenshots / Demo
Not applicable.

## Acceptance criteria
- [x] No separate outgoing/last-message status field is added to `ChatUiState.Ready`.
- [x] Accepted outgoing sends are persisted into the message list with `DeliveryStatus.Sending(0)` before remote send starts.
- [x] The input is cleared after local acceptance, before server acknowledgement, only if the input still matches the submitted text.
- [x] Pre-acceptance validation, business-rule, or local persistence failures keep the input unchanged and do not add a message to the list.
- [x] Local persistence failure does not call the remote send endpoint and surfaces a send error.
- [x] Remote send progress and final statuses update the same persisted message row.
- [x] Remote send failure updates the same persisted message row to `DeliveryStatus.Failed(...)`.
- [x] Failed accepted sends remain visible in the chat message list as failed message bubbles.
- [x] Tests cover local acceptance, clear-input ordering, pre-acceptance failure, remote success updates, and remote failure persistence.

## Breaking Changes
None.

## Follow-ups
- [ ] #389
- [ ] #391


app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageFlowFactory.kt
```diff
@@ -0,0 +1,164 @@
...
+import timur.gilfanov.messenger.data.source.remote.RemoteDataSources
+import timur.gilfanov.messenger.domain.entity.ResultWithError
+import timur.gilfanov.messenger.domain.entity.message.DeliveryError
+import timur.gilfanov.messenger.domain.entity.message.DeliveryStatus
+import timur.gilfanov.messenger.domain.entity.message.Message
+import timur.gilfanov.messenger.domain.entity.message.TextMessage
+import timur.gilfanov.messenger.domain.usecase.common.LocalStorageError
+import timur.gilfanov.messenger.domain.usecase.message.repository.SendMessageRepositoryError
+
+internal class SendMessageFlowFactory(
```
timurgilfanov: Why it's a Factory?

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
```diff
@@ -0,0 +1,164 @@
...
+                                    RemoteDataSourceError.ServerError.toDeliveryError(),
+                                ),
+                            )
+                            emit(
+                                updateFailedMessage(
+                                    failedMessage,
+                                    RemoteDataSourceError.ServerError,
+                                ),
+                            )
+                        } else {
```
timurgilfanov, line 58: Why `uploadedMessage.deliveryStatus == null` means `ServerError` from remote data source?

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatViewModel.kt
```diff
@@ -120,43 +126,75 @@ class ChatViewModel @AssistedInject constructor(
...
+    ) {
+        if (!progress.acceptedLocally && message.deliveryStatus.isAcceptedStatus()) {
+            progress.acceptedLocally = true
+            currentChat = currentChat?.let { chat ->
+                if (chat.messages.none { it.id == message.id }) {
+                    chat.copy(messages = chat.messages.add(message))
+                } else {
+                    chat
+                }
+            }
```
timurgilfanov, line 167: Why we add message to the local cache `currentChat` directly, not waiting for the cache update from repository?

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatViewModel.kt
```diff
@@ -120,43 +126,75 @@ class ChatViewModel @AssistedInject constructor(
...
+            progress.clearInputSent = true
+            currentInputText = ""
+            _effects.send(ChatSideEffect.ClearInputText)
+        }
+    }
+
+    private fun handleSendFailure(error: SendMessageUseCaseError, progress: SendProgress) {
+        _state.update { state ->
+            val showDialog = !progress.acceptedLocally ||
+                error !is SendMessageUseCaseError.RemoteOperationFailed
```
timurgilfanov, line 186: Why we filters out remote data source errors from showing dialog? How there errors will be displayed to the user?

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatScreen.kt
```diff
@@ -224,6 +236,59 @@ fun ChatContent(
...
+            title = { Text(text = stringResource(R.string.send_error_dialog_title)) },
+            text = { Text(text = readyErrorMessage(state.dialogError)) },
+            confirmButton = {
+                TextButton(onClick = onDismissDialogError) {
+                    Text(text = stringResource(R.string.send_error_dialog_dismiss))
+                }
+            },
+            modifier = Modifier.testTag("send_error_dialog"),
+        )
+    }
```
timurgilfanov, line 252: How dialog with error was shown before?

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageFlowFactory.kt
timurgilfanov: It is called `Factory` because this extraction was originally scoped to constructing the send result `Flow` outside `MessengerRepositoryImpl`. After the refinement it also performs the local acceptance side effect before returning the flow, so the name is not perfect. A more precise name would be something like `SendMessageFlowBuilder` or `SendMessageOperation`; there is no architectural requirement for `Factory` specifically.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
timurgilfanov: It does not literally mean the remote data source returned `ServerError`. The repository treats a successful send update with `deliveryStatus == null` as an invalid remote success payload for this send lifecycle: after local acceptance, every remote send emission must carry a delivery status that can be persisted into the same row. Since there is no more specific `RemoteDataSourceError.InvalidPayload` here, `ServerError` is used as the closest remote-operation failure so the row is marked failed and the use case receives a failure instead of silently persisting an unusable status.

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatViewModel.kt
timurgilfanov: The rendered message list still comes from `state.messages` via the repository paging source; this update does not insert directly into the UI list. It only keeps the private `currentChat` snapshot from being stale after local acceptance. That snapshot is used by the ViewModel for local chat context, while the canonical message timeline remains the repository/local data source updates.

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatViewModel.kt
timurgilfanov: Remote failures are filtered out only after the message has already been accepted locally. In that case the failure is displayed through the message row itself as `DeliveryStatus.Failed(...)`, so also showing a dialog would duplicate the same failure and interrupt the composer flow. Remote failures before local acceptance still show the dialog because `progress.acceptedLocally` is false, and local/business-rule failures also show the dialog because there may be no failed message bubble to display.

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatScreen.kt
timurgilfanov: Before this PR, `ChatViewModel` could set `dialogError`, but `ChatScreen` did not render that state, so the user did not actually see a dialog from this screen. This block wires `state.dialogError` to an `AlertDialog` for errors that are not represented by a failed message bubble, mainly pre-acceptance validation/business/local failures.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageFlowFactory.kt
timurgilfanov: It's a wrapper for operation, not a delegation of construction of the object to the method:
```suggestion
internal class SendMessageOperation(
```

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
timurgilfanov: if there a reason for `deliveryStatus` be null? Could we made them a non-nullable in follow-up issue?

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatScreen.kt
timurgilfanov: Is there other screens with `AlertDialog`? Does new dialog match they patterns and texts? Do we need to create a follow-up issue for code reuse for these dialogs?

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
timurgilfanov: Currently `null` is used for the pre-send/draft domain message state: `SendMessageUseCase` explicitly requires outgoing messages to start with `deliveryStatus == null`, then the repository accepts them into the local timeline as `Sending(0)`. Remote DTOs also allow null today, but for the send-response/progress path I do not see a valid reason for the server to return null after local acceptance. A follow-up issue makes sense, but I would scope it carefully: either make delivery status non-null only for persisted/timeline messages, or split draft outgoing message construction from persisted message entities before making the domain property globally non-null.

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatScreen.kt
timurgilfanov: There are auth blocking dialogs in login/signup, and this chat dialog follows the same basic Material3 `AlertDialog` + title/text/confirm-button pattern. The texts are specific to send failures, so I would not block this PR on reuse. A follow-up could be useful if we want a shared app-level error-dialog component, but I would keep it separate because auth dialogs are blocking and route to settings actions, while this chat dialog is dismissible and scoped to composer/send failures.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageFlowFactory.kt
timurgilfanov: Agreed, `SendMessageOperation` describes the role better than `SendMessageFlowFactory`: it coordinates a send operation with local acceptance plus remote status tracking, not just object construction. I can rename it in this PR if you want that cleanup included here.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
timurgilfanov: Created follow-up #389 to make the delivery-status boundary explicit and move toward non-null `DeliveryStatus` for persisted/timeline messages. I also added it to this PR description under Follow-ups.

timurgilfanov (commented):
Looked into implementation part.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
```diff
@@ -0,0 +1,164 @@
...
+        errorCode = DELIVERY_ERROR_UNAUTHORIZED,
+        message = "Unauthorized",
+    )
+
+    is RemoteDataSourceError.UnknownError -> DeliveryError.UnknownError(
+        errorCode = DELIVERY_ERROR_UNKNOWN,
+        message = cause.message,
+    )
+
+    RemoteDataSourceError.UserBlocked -> DeliveryError.RecipientBlocked
```
timurgilfanov: Why we add semantics to UserBlocked here?

timurgilfanov:
Created follow-up issue #391 to track moving send-rule freshness ownership out of `ChatViewModel`, so this PR can keep its scope focused on using the message timeline as the send status source of truth.

app/src/main/java/timur/gilfanov/messenger/ui/screen/chat/ChatViewModel.kt
timurgilfanov: The earlier answer is still the main point: this is not the source for rendering the message list; `state.messages` remains backed by the repository paging source. The extra reason to update `currentChat` immediately is that this private snapshot is passed back into `SendMessageUseCase` for send-rule checks. In particular, the debounce rule looks at the last message from the sender, so waiting for the debounced/asynchronous repository chat update could let a second immediate send evaluate against stale chat context. The `none { it.id == message.id }` guard keeps this local snapshot update idempotent when the repository update arrives later.

app/src/main/java/timur/gilfanov/messenger/data/repository/SendMessageOperation.kt
timurgilfanov: This mapping is for the persisted delivery status, not for expanding the repository/use-case error semantics. After the message has been accepted locally, a remote failure is written back into the same row as `DeliveryStatus.Failed(reason)`, and that reason is what the failed bubble displays. `RemoteDataSourceError.UserBlocked` is the transport/source error; `DeliveryError.RecipientBlocked` is the closest delivery-domain reason we already have for user-facing message status. The repository error path still maps `UserBlocked` as an unexpected send error today, so this PR only gives the accepted message row the best available failure reason. If we want first-class send/use-case semantics for blocked recipients too, I would keep that as a separate error-taxonomy follow-up.

timurgilfanov (commented):
Answered pending review comments.

Issue #398, opened 2026-06-06
[Bug] flowChatUpdates can terminate on a torn Room snapshot


timurgilfanov:
## Bug description

`LocalChatDataSourceImpl.flowChatUpdates()` (`LocalChatDataSourceImpl.kt:150-194`) reads from two independently-invalidated Room flows without a consistent snapshot, exposing a torn-read race.

It `flatMapLatest`es over `chatDao.flowChatWithParticipants(chatId)` and, inside, `map`s over `messageDao.flowLastMessageInChat(chatId)`. The inner message flow re-emits independently of the captured outer `chatWithParticipants` snapshot. A sync `ChatUpdatedDelta` transaction (`LocalSyncDataSourceImpl.applyChatUpdatedDelta`) can atomically add a new participant *and* a message from that new participant. After commit, Room invalidates both table sets, but the inner message flow can re-emit a fresh last-message against the still-stale captured participant set.

When that fresh message's sender is the newly-added participant absent from the captured snapshot, `EntityMappers.toChat` → `toMessage` hits `error("Sender not found for message $id")` (`EntityMappers.kt:189`). The `.catch` at `LocalChatDataSourceImpl.kt:186-194` only handles `SQLiteException` and rethrows everything else, so the per-chat updates flow terminates permanently.

This race was noted during the #396 review (progress log) as a pre-existing pattern in `flowChatUpdates`, untouched by that work.

## Steps to reproduce

1. Open a chat and collect its updates (`flowChatUpdates`).
2. Trigger a sync delta that, in one transaction, adds a new participant to the chat and a message authored by that new participant.
3. Hit the window where the inner `flowLastMessageInChat` re-emits before the outer participant flow catches up.

## Expected behavior

The flow keeps emitting consistent snapshots; transient torn pairs (a message whose sender is not yet in the observed participant set) are skipped rather than crashing the flow.

## Actual behavior

`toMessage` throws `Sender not found`, the non-`SQLiteException` is rethrown by `.catch`, and the flow terminates permanently (chat screen stops receiving updates until re-subscription). Intermittent, write-timing-dependent.

## Suggested fix

[...]

PR #399, opened 2026-06-06
[Refactor] Avoid loading all messages in flowChatList preview query

### LINKED ISSUE #396: [Refactor] Avoid loading all messages in flowChatList preview query
Related to #149 (which enforces the same "don't load full conversations into memory" principle for the chat detail screen). Standalone — no blocking dependency.

### Current state
`LocalChatDataSourceImpl.flowChatList()` (`app/.../data/source/local/LocalChatDataSourceImpl.kt:119`) is backed by `ChatDao.flowAllChatsWithParticipantsAndMessages()`, which returns `ChatWithParticipantsAndMessages`. That Room `@Relation` eagerly loads **every** `MessageEntity` for **every** chat in the list. The only consumer, `EntityMappers.toChatPreview()` (`EntityMappers.kt:143`), uses the messages collection solely to compute the latest one:

```kotlin
val lastMessage = messages.maxByOrNull { it.createdAt }
```

So the full message history of all chats is materialized into memory just to derive one `lastMessage` and `lastActivityAt` per chat.

### Desired state
`flowChatList()` loads only the data a preview needs: chat + participants + the single most-recent message per chat. Mirror the pattern already used by `flowChatUpdates()` (`LocalChatDataSourceImpl.kt:141`), which combines `ChatWithParticipants` (no messages) with a bounded last-message query (`MessageDao.flowLastMessageInChat`, `ORDER BY createdAt DESC LIMIT 1`). After the rewire, `ChatWithParticipantsAndMessages` and the messages-dependent parts of `toChatPreview` become removable.

### Motivation
Memory and CPU cost of the chat list scales with total message history rather than chat count — a list of N chats pulls all their messages on every emission. This contradicts the project's stated goal of not loading full conversations into memory (the same principle #149 enforces for the chat detail screen). The efficient pattern already exists in the same file, so this is an inconsistency, not a new design.

### Affected layer
Data

### Estimated effort
S

### Suggested priority
medium

### Behavior impact
- [x] This refactoring changes no externally observable behavior

### Affected files or modules
- `data/source/local/LocalChatDataSourceImpl.kt` — `flowChatList()`
- `data/source/local/database/dao/ChatDao.kt` — drop/replace `flowAllChatsWithParticipantsAndMessages`; possibly add a "last message per chat" query
- `data/source/local/database/mapper/EntityMappers.kt` — `toChatPreview()`
- `data/source/local/database/entity/ChatWithParticipantsAndMessages.kt` — likely deletable
- Corresponding tests in `EntityMappersTest` / data-source tests

timurgilfanov:
Closes #396

## Summary
Stops `flowChatList()` from eagerly loading every chat's full message history into memory. The preview query now loads only chat + participants + the single most-recent message per chat, mirroring the bounded pattern already used by `flowChatUpdates()`.

## Changes
- Replaced the eager `ChatDao.flowAllChatsWithParticipantsAndMessages()` (`@Relation` over all `MessageEntity`) with a bounded last-message-per-chat query (`MessageDao.flowLastMessagesPerChat`, `ORDER BY createdAt DESC, id DESC LIMIT 1` per chat).
- Rewired `EntityMappers.toChatPreview()` to take an explicit `lastMessage: MessageEntity?` instead of deriving it from a full messages collection.
- Deleted the now-dead `ChatWithParticipantsAndMessages` relation entity and repointed its tests.
- `flowChatList()` combines the chats-with-participants flow and the last-messages flow, attributing each chat its own last message via `associateBy { it.chatId }`.
- Handles torn cross-flow snapshots while still escalating genuine corruption (see below).

## Important Design Decision
The two combined Room flows are invalidated independently, so `combine` can briefly observe a `(fresh last-message, stale participants)` pair after a sync transaction that adds a participant and their first message atomically. Mapping that pair would drive `toMessage` into `error("Sender not found")` and terminate the flow.

A message whose sender is absent from the participant set has two indistinguishable causes at the flow level: a **transient torn read** (DB is actually consistent) or a **real source-of-truth inconsistency** (the DB genuinely holds an orphan message). A blanket skip would silence both — and because a real orphan persists across every read, it would drop *every* emission forever, silently blanking the list. The write path does not prevent orphans (`validateParticipantContract` only enforces the single-current-user rule), so this is a reachable state that must be surfaced.

The discriminator is a single committed snapshot: a torn pair never existed as one committed state, a real orphan does. So:
- **Fast path:** when the combined flow values contain no orphan, map them directly (no DB read).
- **Verify path:** when an apparent orphan is seen, re-read participants + last-messages inside one `withTransaction`. If the orphan is gone it was a transient torn read → emit `Success`; if it persists it's a real inconsistency → emit a handled `Failure(InvalidData)`, keeping the flow alive and the error visible.

This recovers from the concurrency race without reintroducing the eager all-messages load, and does not conflict with the persistence-boundary fail-fast contract — that contract governs *persisted* data integrity, which here is checked against a consistent committed snapshot rather than a torn cross-flow join.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

`LocalChatDataSourceImplTest` (empty-list emission, per-chat last-message attribution, and orphan-sender → `InvalidData` escalation), `MessageDaoTest`, `ChatDaoTest`, and `EntityMappersTest` pass; `ktlintFormat`/`detekt` clean; `:app:compileMockDebugAndroidTestKotlin` compiles.

## Acceptance criteria
- [x] `flowChatList()` loads only chat + participants + the single most-recent message per chat (no full message history materialized).
- [x] `ChatWithParticipantsAndMessages` and the messages-dependent parts of `toChatPreview` removed.
- [x] No externally observable behavior change in the common case.

## Breaking Changes
None.

## Follow-ups
- [x] #398 — `flowChatUpdates` has the same torn-snapshot race and needs the same treatment (out of scope here).
- [x] #400 — benchmark chat list preview read performance to quantify the last-message scan cost.
- [x] #401 — add a composite `(chatId, createdAt, id)` index to fix the last-message-per-chat quadratic scan.


app/src/test/java/timur/gilfanov/messenger/data/source/local/database/dao/ChatDaoTest.kt
```diff
@@ -165,14 +199,15 @@ class ChatDaoTest {
         messageDao.insertMessage(message)
         chatDao.insertChatParticipantCrossRef(crossRef)
 
-        val result = chatDao.getChatWithParticipantsAndMessages(chatId)
+        val result = chatDao.getChatWithParticipants(chatId)
+        val lastMessage = messageDao.getLastMessageInChat(chatId)
```
timurgilfanov: Why we test `messageDao` here?

docs/plans/completed/2026-06-05-avoid-loading-all-messages-in-flowchatlist.md
```diff
@@ -0,0 +1,91 @@
...
+  - `app/src/main/java/timur/gilfanov/messenger/data/source/local/database/mapper/EntityMappers.kt` — `toChatPreview()`
+  - `app/src/main/java/timur/gilfanov/messenger/data/source/local/database/entity/ChatWithParticipantsAndMessages.kt` — delete
+  - `app/src/main/java/timur/gilfanov/messenger/data/source/local/database/entity/ChatWithParticipants.kt` — reference for `toChat(lastMessages)` pattern
+  - Tests:
+    - `app/src/test/java/timur/gilfanov/messenger/data/source/local/LocalChatDataSourceImplTest.kt`
+    - `app/src/test/java/timur/gilfanov/messenger/data/source/local/database/mapper/EntityMappersTest.kt`
+    - `app/src/test/java/timur/gilfanov/messenger/data/source/local/database/dao/ChatDaoTest.kt`
+    - `app/src/test/java/timur/gilfanov/messenger/data/source/local/database/dao/ParticipantDaoTest.kt`
+    - `app/src/test/java/timur/gilfanov/messenger/data/source/local/database/dao/MessageDaoTest.kt`
+- Related patterns: `flowChatUpdates()` + `MessageDao.flowLastMessageInChat` (createdAt DESC, id DESC LIMIT 1); existing `ChatWithParticipants` entity and `ChatWithParticipants.toChat(lastMessages)` mapper.
```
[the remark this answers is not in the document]
timurgilfanov: False positive — `flowLastMessageInChat` was not removed. It still exists in `MessageDao` (line 55) and is used by `flowChatUpdates()` (`LocalChatDataSourceImpl.kt:220`). Line 20 documents the *existing single-chat* related pattern that the new list flow is modeled after; the new plural method `flowLastMessagesPerChat()` is a separate addition for the list query. The reference is correct as written, so no change.

app/src/test/java/timur/gilfanov/messenger/data/source/local/database/dao/ChatDaoTest.kt
timurgilfanov: Good catch — it shouldn't. This is a leftover from the old combined test of `getChatWithParticipantsAndMessages`: when that entity was split, the message half was awkwardly kept here by calling `messageDao.getLastMessageInChat`. That method is already covered by `MessageDaoTest` (`get last message in chat returns most recent message`, line 149), so the assertion here was redundant and out of place in a `ChatDao` test.

Fixed: removed the message insert and `messageDao.getLastMessageInChat` assertions, so the test now focuses solely on `chatDao.getChatWithParticipants`.

PR #331, opened 2026-04-06
feat: auth-gated navigation and feature integration

timurgilfanov:
Closes #274

## Summary
Wires all Phase 1 auth verticals together: replaces the hardcoded identity stub (`DefaultIdentityRepository`) with `AuthRepository`-backed identity, connects Settings and Chat to live auth state, and adds auth-gated startup/logout navigation.

## Changes
- Add `isCurrentUser` field to `Participant` entity to determine authenticated user. It allow us to remove stable user ID as requested in Specification.
- Scope settings by `UserScopeKey` (SHA-256 of refresh token) at the `SettingsRepository` domain boundary
- Remove `DefaultIdentityRepository` and `IdentityRepository`; inject `AuthRepository` directly into settings and profile use cases
- Add `MainActivityViewModel` to observe auth state and emit navigation side-effects (startup routing + logout redirect)

- Add Room auto-migration with settings table clear on migration from schema 4 to 5
- Add end-to-end `AuthApplicationTest` covering login → chat → logout → login flow

## Important Design Decision
`UserScopeKey` is derived from the refresh token hash rather than a stable user ID so it rotates on each session, isolating settings data per session without exposing raw tokens to the domain layer. This intentional design has a known follow-up: see #330.

## Testing
- Unit tests added/updated for all affected use cases and ViewModels (`MainActivityViewModelTest`, `ChatViewModelAuthTest`, `SyncSettingUseCaseTest`, etc.)
- End-to-end `AuthApplicationTest` validates the full login → chat → logout → login flow on a real emulator
- Ignore `MessageReplyApplicationTest` because PR reveals problems that out of scope of this PR: 
  - #338
  - #339
Test must be returned after those problems fixed.

## Acceptance criteria
- [x] `DefaultIdentityRepository` removed from production DI
- [x] All identity references use auth-backed identity
- [x] App startup routes to correct screen based on auth state
- [x] Auth state changes trigger appropriate navigation
- [x] Settings displays authenticated user info
- [x] Chat attributes messages to authenticated user
- [x] No hardcoded `currentUserId` remains in codebase
- [x] End-to-end flow works: login → chat → logout → login

## Follow-ups
- #330
- #338
- #339
- #335

PR #402, opened 2026-06-06
Implement profile data layer

### LINKED ISSUE #336: [Feature] Implement profile data layer
## Problem statement

Profile data is currently represented by an intentional stub because the profile data layer has not been implemented yet. The application needs a dedicated profile data layer so profile information can be sourced through the intended architecture instead of placeholder values.

## Proposed solution

Implement the profile data layer for profile retrieval and expose it through the existing domain and use case contracts. Replace the temporary stub in `ObserveProfileUseCaseImpl` with data backed by the new profile data layer once that layer exists.

## Alternatives considered

Keep the hardcoded stub indefinitely. This was rejected because the stub is only a temporary placeholder and does not satisfy the intended product behavior.

## Affected layer

Data

## Suggested priority

high

## Estimated effort

M

## Additional context

This issue reclassifies the current behavior as planned implementation work rather than a bug. The existing hardcoded profile is an intentional temporary stub until the profile data layer is available.

timurgilfanov:
Closes #336

## Summary
Implements the profile retrieval data layer so the profile screen uses backend-backed data instead of the temporary hardcoded profile.

## Changes
- Add Room profile persistence, schema migration, and local data source.
- Add the Ktor remote profile source and cache-then-remote repository implementation.
- Replace the observe-profile use case stub and surface loading and retrieval failures in the profile UI.
- Clear profile and settings data through `UserScopedDataCleaner` implementations during logout.
- Add focused repository, data source, use case, ViewModel, and cleanup tests.

## Important Design Decision
Room is the local source of truth for profile observation: cached data is emitted first, backend responses are written to Room, and subsequent updates continue through the Room flow.

Future profile reads and writes should preserve this model. New user-scoped storage must implement `UserScopedDataCleaner` and join the existing multibinding so logout cleanup remains centralized. Profile editing should extend the repository without replacing the observation model.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

## Screenshots / Demo
Not applicable; this PR changes profile data loading and error handling without introducing a new visual design.

## Acceptance criteria
- [x] The profile screen retrieves the current profile through the profile data layer instead of a hardcoded stub.
- [x] Profile data is cached locally and refreshed from the backend.
- [x] User-scoped profile data is removed during logout.

## Breaking Changes
None.

## Follow-ups
- [ ] Implement profile name and picture editing in a separate vertical slice.


timurgilfanov:
@coderabbitai review

timurgilfanov:
@coderabbitai help

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserverImpl.kt
```diff
@@ -65,9 +65,11 @@ class AuthCleanupObserverImpl @Inject constructor(
...
 
     private suspend fun deleteUserData(userKey: UserScopeKey) {
-        val result = settingsRepository.get().deleteUserData(userKey)
-        if (result is ResultWithError.Failure) {
-            logger.e(TAG, "Failed to clean up user-scoped settings: ${result.error}")
+        dataCleaners.get().forEach { cleaner ->
+            val result = cleaner.deleteUserData(userKey)
+            if (result is ResultWithError.Failure) {
+                logger.e(TAG, "Failed to clean up user-scoped data: ${result.error}")
+            }
```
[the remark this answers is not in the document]
timurgilfanov: Fixed in f85e924f: each cleaner call is now wrapped in try/catch (rethrowing CancellationException, logging other throwables) so one throwing cleaner can no longer cancel the observer.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -89,13 +90,35 @@ class AuthRepositoryImpl @Inject constructor(
...
+            if (sessionIdWasMissing) {
+                sessionMutex.withLock {
+                    localDataSource.saveSession(session).fold(
+                        onSuccess = {},
+                        onFailure = { e ->
+                            logger.e(TAG, "Failed to persist generated session ID: $e")
+                        },
+                    )
+                }
+            }
```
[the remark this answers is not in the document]
timurgilfanov: Fixed in f85e924f: restoreAuthState now runs entirely under sessionMutex (reads + saveSession + compareAndSet), so a concurrent login/signup either wins the lock first (restore then reads the fresh session and the CAS no-ops) or waits, eliminating the stale-snapshot overwrite.

core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/entity/auth/AuthSession.kt
```diff
@@ -1,3 +1,8 @@
 package timur.gilfanov.messenger.domain.entity.auth
 
-data class AuthSession(val tokens: AuthTokens, val provider: AuthProvider)
+/**
+ * @property sessionId Stable identifier for the logical login session. Generated at login/signup
+ *   and preserved across token rotations so that user-scoped data partitions remain stable when
+ *   the refresh token is rotated.
+ */
+data class AuthSession(val tokens: AuthTokens, val provider: AuthProvider, val sessionId: String)
```
timurgilfanov, line 8: Why we not use `AuthTokens.refreshToken` as sessionId as we used it before?
```
fun AuthSession.toUserScopeKey(): UserScopeKey = UserScopeKey(tokens.refreshToken.sha256Hex())
```
Refresh token will be updated only on logout and new login and will not be rotated during session life-time? Also, what the trade-offs the design with non-rotating refreshToken comparing to rotating.

core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/usecase/profile/repository/ProfileRepository.kt
```diff
@@ -16,12 +16,14 @@ import timur.gilfanov.messenger.domain.usecase.profile.ObserveProfileError
...
-     * Observes profile changes for a specific user.
+     * Observes the authenticated user's global profile.
      *
-     * @param userId The unique identifier of the user to observe
+     * @param userKey Opaque local key identifying the authenticated user's cache partition
      * @return Flow emitting profile updates or errors
      */
-    fun observeProfile(userId: UserId): Flow<ResultWithError<Profile, ObserveProfileError>>
+    fun observeCurrentProfile(
+        userKey: UserScopeKey,
```
timurgilfanov, line 25: Why we change UserId to UserScopeKey here? Could you also remind me a motivation for introduction of the UserScopeKey and. I remember we introduce UserScopeKey for something like preference stored per user.

core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/entity/auth/AuthSession.kt
timurgilfanov: The premise needs one correction: the **refresh token *is* rotated during the session's lifetime**, not only on logout/new login.

`AuthRepositoryImpl.refreshToken()` calls `remoteDataSource.refresh(...)` and persists the returned `newTokens` via `saveRefreshedTokens` — which does `saveTokens(newTokens)` and updates `_authState` with `session.copy(tokens = newTokens)`. So every time the access token expires and we refresh, a new refresh token replaces the old one.

If `toUserScopeKey()` were `sha256(refreshToken)`, the scope key would change on every routine refresh, and all per-user **local** data partitioned by it (settings/preferences, now the profile cache) would silently become unreachable mid-session — the user would appear to "lose" cached data after a background refresh, and `AuthCleanupObserver` would even treat the refresh as a user-scope change and wipe the old partition.

That is why `sessionId` exists: generated once at login/signup and preserved across token rotations (see the KDoc on `AuthSession.sessionId`), so the partition is stable for the whole login session and only changes on logout → new login (a genuinely new scope). `sha256(refreshToken)` is exactly what `Migration5To6` migrates away from.

**Rotating vs non-rotating trade-off:**
- *Rotating* (what we do) is the security best practice — each refresh invalidates the previous refresh token, shrinking the window a stolen token is usable and enabling reuse-detection / family revocation on replay. Cost: the refresh token is no longer a stable identifier, so it cannot double as a scope key.
- *Non-rotating* would let the refresh token serve as a stable key, but a leaked refresh token stays valid for the whole session with no reuse detection — weaker security.

So we deliberately decoupled the rotating security credential (refresh token) from the stable scoping identity (sessionId).

core/domain/src/main/kotlin/timur/gilfanov/messenger/domain/usecase/profile/repository/ProfileRepository.kt
timurgilfanov: `UserScopeKey` is an opaque **local** partition key for user-scoped on-device data, derived as `sha256(sessionId)`. Your memory is right: it was introduced with the auth integration (#331) for per-user **settings/preferences**, and the profile cache is the same class of data, so it reuses the same mechanism.

Why `UserScopeKey` rather than `UserId` for `observeCurrentProfile`:

1. **There is no `UserId` available in this path.** `observeCurrentProfile` reads/writes the **local cache** and refreshes from `remoteDataSource.getCurrentProfile()`, which takes no id — the server resolves the current user from the access token. And `Profile` itself is just `name` + `pictureUrl`, no id. So there is no server `UserId` to key the local cache on; what we do have, the moment we are authenticated, is the session → `UserScopeKey`.

2. **Local partitioning decoupled from server identity.** The lifecycle of cached data is tied to the login *session* — `UserScopedDataCleaner` wipes a partition when the scope key changes (logout / new login). The correct boundary for on-device caches/preferences is "this device's logged-in session," not a server user id.

3. **Opacity.** It is a hash, so the on-disk partition key exposes neither the raw sessionId nor any PII.

Note the deliberate split in this interface: `observeCurrentProfile` takes `UserScopeKey` because it reads the **local cache partition**, while `updateName` / `updatePicture` / `removePicture` keep `UserId` because they target the **remote server identity**. Local partition key vs server identity — different concerns, different types.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
timurgilfanov, line 1: New session id generation should be extacted to private function for reuse.

```
private fun generateSessionId() = UUID.randomUUID().toString()
```

Should we use `inline` here?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserverImpl.kt
```diff
@@ -4,21 +4,22 @@ import dagger.Lazy
...
 import timur.gilfanov.messenger.domain.toUserScopeKey
 import timur.gilfanov.messenger.domain.usecase.auth.AuthRepository
-import timur.gilfanov.messenger.domain.usecase.settings.repository.SettingsRepository
 import timur.gilfanov.messenger.util.Logger
 
 @Singleton
 class AuthCleanupObserverImpl @Inject constructor(
     private val authRepository: Lazy<AuthRepository>,
-    private val settingsRepository: Lazy<SettingsRepository>,
+    private val dataCleaners: Lazy<Set<@JvmSuppressWildcards UserScopedDataCleaner>>,
```
timurgilfanov, line 26: Why @JvmSuppressWildcards?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -89,13 +90,33 @@ class AuthRepositoryImpl @Inject constructor(
                 null
             },
         )
+        val getSessionIdResult = localDataSource.getSessionId()
```
timurgilfanov: If we failed on read `sessionId` or write `AuthSession` locally what can goes wrong?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -67,7 +68,7 @@ class AuthRepositoryImpl @Inject constructor(
         }
     }
 
-    private suspend fun restoreAuthState() {
+    private suspend fun restoreAuthState() = sessionMutex.withLock {
```
timurgilfanov, line 71: Why lock this block?

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserverImpl.kt
```diff
@@ -65,9 +66,21 @@ class AuthCleanupObserverImpl @Inject constructor(
...
+                }
+            } catch (cancellation: CancellationException) {
+                throw cancellation
+            } catch (@Suppress("TooGenericExceptionCaught") cause: Throwable) {
+                logger.e(
+                    TAG,
+                    "Cleaner threw while deleting user-scoped data for $userKey",
+                    cause,
+                )
+            }
```
timurgilfanov: Should rely on returning error. Exception should cause crash.
```suggestion
        dataCleaners.get().forEach { cleaner ->
            cleaner.deleteUserData(userKey).fold(            
                onFailure = { error ->
                    switch(error) {
                        e1 -> error1
                    }
                }
            )
        }
```

deleteUserData should return result with error and callers should handle it.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/source/local/LocalAuthDataSourceImpl.kt
```diff
@@ -164,6 +155,22 @@ class LocalAuthDataSourceImpl internal constructor(
     )
 }
 
+private fun AuthSessionCipher.encryptTokenPair(
```
timurgilfanov: Why made it `AuthSessionCipher` extension function?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/data/repository/ProfileRepositoryImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+        when (val remoteResult = remoteDataSource.getCurrentProfile()) {
+            is ResultWithError.Success -> {
+                val upsertResult = localDataSource.upsertProfile(
+                    userKey = userKey,
+                    profile = remoteResult.data,
+                )
+                if (upsertResult is ResultWithError.Failure) {
+                    logger.w(TAG, "Failed to cache profile locally: ${upsertResult.error}")
+                }
+                emit(ResultWithError.Success(remoteResult.data))
```
timurgilfanov: We should not emit remote date if it's not stored to local data source, because local data source is SOT.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/data/source/remote/RemoteProfileDataSource.kt
timurgilfanov, line 1: Why interfaces other then `getCurrentProfile` removed?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/domain/usecase/ObserveProfileUseCaseImpl.kt
timurgilfanov, line 1: Why move `flatMapLatest` from `AuthState` to `userScopeKey`?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
timurgilfanov, line 1: Why we don't provide more specific error so user could have recovery path for some of errors and more specific information on errors?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
```diff
@@ -29,33 +32,73 @@ class ProfileViewModel @Inject constructor(
...
+                            is ObserveProfileError.RemoteOperationFailed -> {
+                                logger.i(
+                                    TAG,
+                                    "Profile observation remote failure: ${error.error}",
+                                )
+                                if (_state.value is ProfileUiState.Loading) {
+                                    _state.value = ProfileUiState.Error
+                                    savedStateHandle[KEY_ERROR_STATE] = true
+                                }
+                                _effects.send(Feedback.ObserveFailed)
```
timurgilfanov: Why we need to send feedback side-effect even we change a state to `Error`?

timurgilfanov (commented):
First pass completed.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserverImpl.kt
timurgilfanov: Done in f9990be3. Removed the generic `try/catch`: `deleteUserData` now relies on the cleaner's `ResultWithError` and aggregates best-effort failures, which `handleTransition` logs. A genuinely throwing cleaner is no longer caught — it propagates and fails fast, per our error-handling convention.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
timurgilfanov: Done in c4aec44b. Extracted `private fun generateSessionId()` and replaced the four login/signup sites. Kept it a regular function — `inline` gives no benefit here (no lambda or reified type parameters).

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthRepositoryImpl.kt
timurgilfanov: Good catch — treating a read failure like "missing" regenerated and persisted a new session ID, clobbering a recoverable one and orphaning user-scoped data. Fixed in 1cf1b085: `restoreAuthState` now treats a null or failed session-ID read as non-restorable and yields `Unauthenticated`, and no longer generates or persists a session ID. The re-login snackbar UX is tracked in #415.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/data/repository/ProfileRepositoryImpl.kt
timurgilfanov: Agreed. Fixed in 431ebf07: on remote success, if the local upsert fails, `observeCurrentProfile` now emits `Failure(LocalOperationFailed)` instead of the un-persisted remote profile, keeping the local source of truth authoritative.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
timurgilfanov: Fixed in 4d7df682. Both failure branches now go through `handleObserveFailure()`, which posts `Feedback.ObserveFailed` only while in `Ready` (cached data on screen). On `Loading → Error` and when already in `Error` the toast is suppressed — the Error screen already communicates failure.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/repository/AuthCleanupObserverImpl.kt
timurgilfanov: Documented in 48eb1a30. Dagger multibindings provide an invariant `Set<UserScopedDataCleaner>`, but Kotlin would otherwise emit `Set<? extends UserScopedDataCleaner>`, which the binding can't resolve — `@JvmSuppressWildcards` keeps it invariant.

feature/auth/src/main/kotlin/timur/gilfanov/messenger/auth/data/source/local/LocalAuthDataSourceImpl.kt
timurgilfanov: Changed in fb6f1ecb to a private member function that uses the `authSessionCipher` field directly — it's only used inside this class, so the extension receiver added nothing. The extra member trips detekt `TooManyFunctions` (11), so the class now carries `@Suppress("TooManyFunctions")`.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
timurgilfanov: Good point — the single `ProfileUiState.Error` collapses the distinct `ObserveProfileError` cases. Tracked as a follow-up in #414 (carry the error variant into UI state for per-error messaging and recovery); leaving it out of this data-layer PR.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/domain/usecase/ObserveProfileUseCaseImpl.kt
timurgilfanov: `toUserScopeKey()` is `UserScopeKey(sessionId.sha256Hex())`, and the session ID is stable across token refresh (only the access/refresh tokens rotate). So mapping `authState → userScopeKey` and applying `distinctUntilChanged` before `flatMapLatest` means the downstream `observeCurrentProfile` flow is torn down and restarted only when the user scope actually changes (login / logout / account switch).

If `flatMapLatest` keyed on the raw `AuthState` instead, every token refresh emits a new `AuthState` (new `AuthSession`), which would cancel and restart the profile observation — re-fetching the remote profile and re-subscribing to the local cache on each refresh. The mapped key avoids that churn, and it's also the cache partition key `observeCurrentProfile` needs anyway.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/data/source/remote/RemoteProfileDataSource.kt
timurgilfanov: Two reasons:

1. The edit operations (`updateName`, `updatePicture`, `removePicture`) are out of scope for this PR, which delivers only the profile *retrieval/observe* vertical. The repository's edit methods deliberately return `editOutOfScopeError()` ("Profile edit operations are out of scope for profile retrieval"); the remote edit methods will be reintroduced together with the profile-edit vertical (vertical feature delivery).
2. `getProfile(userId)` became `getCurrentProfile()`: the backend returns the authenticated user's own profile, identified by the auth token, so no `userId` parameter is needed. On-device scoping is handled by `UserScopeKey` in the repository rather than a remote id.

timurgilfanov: Correction to my earlier reply — "out of scope" was the wrong justification for *deleting* these methods: the deletion is itself a reviewable change, and it left the three edit error types (`UpdateNameRemoteDataSourceError`, `UpdatePictureRemoteDataSourceError`, `RemovePictureRemoteDataSourceError`) orphaned while the repository layer kept the edit contract (stubbed with `editOutOfScopeError()`).

Reverted in 12efeefb: `updateName`/`updatePicture`/`removePicture` are restored on `RemoteProfileDataSource` and implemented in `RemoteProfileDataSourceImpl` as error-returning stubs (out-of-scope), consistent with the repository's existing pattern — out-of-scope work gets an error implementation rather than a deletion. The only retrieval-related change that remains is `getProfile(userId)` → `getCurrentProfile()`. Real backend wiring for the edit ops arrives with the profile-edit vertical.

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/data/repository/ProfileRepositoryImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+            is ResultWithError.Failure -> when (firstLocal.error) {
+                ProfileLocalDataSourceError.UserDataNotFound -> Unit
+                is ProfileLocalDataSourceError.LocalDataSource -> {
+                    emit(ResultWithError.Failure(firstLocal.error.toObserveProfileError()))
+                    return@flow
+                }
+            }
+        }
+
+        when (val remoteResult = remoteDataSource.getCurrentProfile()) {
```
timurgilfanov, line 64: Should we request remote data source for profile on each profile observation? May be some cache required to prevent fetching data while user just browsing through the app?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
```diff
@@ -29,33 +32,75 @@ class ProfileViewModel @Inject constructor(
 
     private val _state = MutableStateFlow(
         savedStateHandle.get<ProfileUi>(KEY_PROFILE)?.let { ProfileUiState.Ready(it) }
-            ?: ProfileUiState.Loading,
+            ?: if (savedStateHandle.get<Boolean>(KEY_ERROR_STATE) == true) {
+                ProfileUiState.Error
+            } else {
+                ProfileUiState.Loading
+            },
```
timurgilfanov, line 39: Should we check for the error first, and then for the profile?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
```diff
@@ -29,33 +32,75 @@ class ProfileViewModel @Inject constructor(
...
+                            }
+
+                            is ObserveProfileError.LocalOperationFailed ->
+                                handleObserveFailure()
+
+                            is ObserveProfileError.RemoteOperationFailed -> {
+                                logger.i(
+                                    TAG,
+                                    "Profile observation remote failure: ${error.error}",
+                                )
```
timurgilfanov, line 78: Why remote have log and local not?

feature/profile/src/main/kotlin/timur/gilfanov/messenger/profile/ui/screen/ProfileViewModel.kt
```diff
@@ -29,33 +32,75 @@ class ProfileViewModel @Inject constructor(
...
+                savedStateHandle[KEY_ERROR_STATE] = true
+            }
+            ProfileUiState.Error -> Unit
+        }
+    }
+
+    fun retry() {
+        if (_state.value is ProfileUiState.Error) {
+            _state.value = ProfileUiState.Loading
+            savedStateHandle.remove<Boolean>(KEY_ERROR_STATE)
```
timurgilfanov, line 102: Should we remove profile too?

PR #404, opened 2026-06-06
Fix flowChatUpdates torn Room snapshots

### LINKED ISSUE #398: [Bug] flowChatUpdates can terminate on a torn Room snapshot

timurgilfanov:
Closes #398

## Summary
Prevents `flowChatUpdates` from terminating when independently invalidated Room flows temporarily expose a fresh last message with a stale participant snapshot.

## Changes
- Detect apparent orphan last-message senders before mapping a chat update.
- Verify apparent orphans by re-reading participants and the last message in one Room transaction.
- Recover transient torn reads while surfacing persistent orphan data as handled `InvalidData` failures.
- Make last-message selection deterministic for equal timestamps with an `id DESC` tie-breaker.
- Add component tests for the normal path, persistent-orphan failure and recovery, and deterministic DAO ordering.

## Important Design Decision
The flow uses a fast path for already-consistent snapshots and only performs a transactional re-read when the captured participant set does not contain the last-message sender. This avoids an extra database read for normal emissions while distinguishing a transient cross-flow torn read from a real persisted inconsistency. Persistent orphans remain visible as handled errors instead of being silently skipped.

## Testing
- [ ] Manually tested
- [x] Edge cases considered

## Screenshots / Demo
Not applicable; no UI changes.

## Acceptance criteria
- [x] `flowChatUpdates` keeps emitting consistent snapshots instead of terminating on a transient torn participant/message pair.
- [x] A persistent last-message sender orphan is surfaced as a handled `LocalDataSourceError.InvalidData`.
- [x] The same flow subscription can recover and emit success after invalid data is repaired.
- [x] Normal chat updates continue to use the direct mapping path.

## Breaking Changes
None.

## Follow-ups
None.


app/src/main/java/timur/gilfanov/messenger/data/source/local/LocalChatDataSourceImpl.kt
```diff
@@ -191,6 +194,32 @@ class LocalChatDataSourceImpl @Inject constructor(
...
+     *   participants and last message inside one `withTransaction`. If the orphan is gone it was a
+     *   transient torn read and emits [ResultWithError.Success]; if it persists it is a real
+     *   inconsistency and emits a handled [LocalDataSourceError.InvalidData] [ResultWithError.Failure],
+     *   keeping the flow alive and the error visible. Escalation (not a silent skip) is correct
+     *   because a real orphan would otherwise blank the chat forever.
+     *
+     * Note: with a single real Room DB the verify path's transient-recovery branch is not
+     * deterministically forceable through this public flow — a genuinely consistent DB always
+     * takes the fast path, and the only way to observe a persisting orphan is to seed a real one,
+     * which escalates. The recovery branch is therefore not unit-forceable here.
```
timurgilfanov: Do we need this comment?
```suggestion
```

app/src/main/java/timur/gilfanov/messenger/data/source/local/LocalChatDataSourceImpl.kt
```diff
@@ -191,6 +194,32 @@ class LocalChatDataSourceImpl @Inject constructor(
...
+     *
+     * To discriminate without silencing real orphans:
+     * - **Fast path**: when the captured participant snapshot already contains the last message's
+     *   sender, map directly with no extra DB read.
+     * - **Verify path**: when an apparent orphan is seen, [readChatUpdateConsistently] re-reads
+     *   participants and last message inside one `withTransaction`. If the orphan is gone it was a
+     *   transient torn read and emits [ResultWithError.Success]; if it persists it is a real
+     *   inconsistency and emits a handled [LocalDataSourceError.InvalidData] [ResultWithError.Failure],
+     *   keeping the flow alive and the error visible. Escalation (not a silent skip) is correct
+     *   because a real orphan would otherwise blank the chat forever.
```
timurgilfanov: Do we need this much details here?

